PR #189, opened 2020-02-27
Memory generator REST API

abelov-spirent:
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/189)


abelov-spirent:
@dmorton-spirent Hello
```
> swagger-cli validate memory.yaml 
memory.yaml is valid
```

PR #195, opened 2020-03-07
modules/socket: Added support for IPv6

cmyers-spirent:
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/195)


api/schema/v1/modules/packetio.yaml
```diff
@@ -332,6 +337,7 @@ definitions:
             enum:
               - static
               - dhcp
+              - auto
```
cmyers-spirent: Added explicit configuration method for "auto" instead of using missing ipv4 interface to indicate it.
This allows for configuring IPv6 only stacks.
IPv6 config has same "auto" option for SLAAC.

api/schema/v1/modules/packetio.yaml
```diff
@@ -361,6 +367,49 @@ definitions:
...
+                type: string
+                description: Default gateway
+            required:
+              - address
+              - prefix_length
+          dhcp6:
+            type: object
+            description: DHCPv6 configuration parameters
+            properties:
+              stateless:
```
cmyers-spirent: Setting "stateless" to false currently cause an error because lwip built-in DHCPv6 doesn't support stateful mode.

api/schema/v1/modules/packetio.yaml
```diff
@@ -361,6 +367,49 @@ definitions:
...
+              address:
+                type: string
+                description: IPv6 address
+              prefix_length:
+                type: integer
+                description: Prefix length
+                format: int32
+              gateway:
+                type: string
+                description: Default gateway
```
cmyers-spirent: IPv6 prefix length and gateway options are not supported by lwip.
Prefix lengths are always /64 and gateway is not supported because it only supports routes/prefixes from NDP.

src/modules/socket/server/icmp_socket.cpp
```diff
@@ -37,8 +42,10 @@ icmp_receive_raw(void* arg, raw_pcb*, pbuf* p, const ip_addr_t* addr)
     auto sock = reinterpret_cast<icmp_socket*>(arg);
 
     if (sock->is_filtered(icmp_type(p))) {
-        pbuf_free(p);
-        return (1); /* pbuf was consumed */
+        /* packet not handled by socket so return 0 to allow
+         * icmp_input()/icmp6_input() to handle it
+         */
+        return 0;
```
cmyers-spirent: Returning 1 was causing the packets to not get handled by icmp6/ndp so it seems we should return 0 for packets not handled by the socket.

src/modules/socket/server/raw_socket.cpp
```diff
@@ -16,6 +16,17 @@ namespace openperf {
...
+ * structures must have exactly the same memory layout
+ */
+static_assert(sizeof(struct dgram_ipv4_addr) == sizeof(ip4_addr_t));
+static_assert(sizeof(struct dgram_ipv6_addr) == sizeof(ip6_addr_t));
+static_assert(sizeof(((struct dgram_ip_addr*)0)->type)
+              == sizeof(((ip_addr_t*)0)->type));
+static_assert(offsetof(struct dgram_ip_addr, type)
+              == offsetof(struct ip_addr, type));
+static_assert(sizeof(struct dgram_ip_addr) == sizeof(ip_addr_t));
+
```
cmyers-spirent: Added a bunch of static_asserts() so we can tell if the strucutres don't match up again.

tests/aat/spec/dataplane_spec.py
```diff
@@ -302,8 +308,8 @@ def do_ping(api_client, ping_binary, src_id, dst_id, domain):
         self.process = service.start()
         self.api = client.api.InterfacesApi(service.client())
 
-    with description('ipv4,'):
-        with description('ping,'):
+    with description('ipv4,', 'dataplane:ipv4'):
+        with description('ping,', 'dataplane:ipv4:ping'):
```
cmyers-spirent: mamba -t <tag>,<tag> seems to be logical OR.  so using ":" to explicitly define tag hiearchy.
mamba -t dataplane  <- Runs all dataplane tests
mamba -t dataplane:ipv4 <- Runs all dataplane IPv4 tests
mamba -t dataplane:ipv4:ping  <- Runs all dataplane IPv4 ping tests

If there is a better way to do this then LMK

tests/aat/spec/dataplane_spec.py
```diff
@@ -458,9 +464,170 @@ def do_ping(api_client, ping_binary, src_id, dst_id, domain):
...
+                        expect(client_rx_stop - client_rx_start) \
+                            .to(be_above_or_equal(read))
+                        expect(server_tx_stop - server_tx_start) \
+                            .to(be_above_or_equal(written))
+
     with after.all:
         try:
-            self.process.terminate()
-            self.process.wait()
+            if self.process:
```
cmyers-spirent: Ooops.  this wasn't supposed to get checked in...

PR #208, opened 2020-03-25
Block I/O load generator

abelov-spirent:
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/208)


loskutnikov-spirent:
> Identifiers with leading underscores are reserved per the C++ standard. Please do not use them. Most openperf code uses the prefix m_ for member variables, but I understand the cool kids use trailing underscores, e.g. name_.

Identifiers with double underscores are reserved per the C++ standard anywhere. Identifiers with single leading underscore reserved in global namespace scope only, so it securely and allowed for using inside the class/struct for naming private and protected members in our case. Besides, _name more readable and visually distinguished relatively to name_.

loskutnikov-spirent:
Well, if we will discuss about "potentiality" or "possibility" can conclude we cannot exclude overlapping or shadowing any other global identifier in any other scope. Fortunately, C++ creators foresee and gave us name scopes and qualified names to prevent it - it is good instrument. If you want use identifier from global scope, simply use ::\_name and be happy. Also, it give you more information than simply \_name and emphase that you use global \_name. I think, dont worth introduce limitations where they are not (according to C++ standard).

abelov-spirent:
___
*[src/framework/utils/worker/worker.hpp, line 81 at r3](https://reviewable.io/reviews/spirent/openperf/208#-M3HtRxi2ek4oual4-lL:-M5N1f4F0Ar8B2X4dWAt:bho7t6o) ([raw file](https://github.com/spirent/openperf/blob/0fe059b2473a0613e82ac76010fd7e9521e319ec/src/framework/utils/worker/worker.hpp#L81)):*

The main 0MQ context is also used to signal a clean shutdown.  How does this generator know when it is time to shut itself down?
</blockquote></details>

Block service and block server are still using main 0MQ context. Local context is created just to transfer data between main thread and worker thread.

PR #245, opened 2020-06-04
Dynamic Results REST API

loskutnikov-spirent:
REST API specification for future "Dynamic" module.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/245)


loskutnikov-spirent:
> > _[Reviewable](https://reviewable.io/reviews/spirent/openperf/245)_ status: 2 of 31 files reviewed, 1 unresolved discussion (waiting on @djoyner)
> 
> _[a discussion](https://reviewable.io/reviews/spirent/openperf/245#-M9yWolN2coacJ1wPx00:-MA7wf7YCzph2J9P-lHn:b-50nzvj) (no related file):_
> Previously, jhiga (Justin Higa) wrote…
> 
> Agree dynamic results are not generators. However, dynamic results are at their core an interpretation of a result set. As a result, they are intrinsically tied to the result set that generated them. By making thresholds and t-digests part of a generator configuration and result, that link is explicit and the lifetimes of the objects are strongly coupled. This prevents any issues where one exists without the other.
> 
> Making dynamic results top level entities increases the burden on clients in order to use them. For example, let's say a client wants to run a test with both threshold and t-digest results. The series of operations becomes:
> 
>     1. POST /generator
> 
>     2. POST /threshold
> 
>     3. POST /t-digest
> 
>     4. POST /generator/{gen_id}/start
> 
>     5. POST /generator/stop
> 
>     6. GET /generator-result/{result_id}
> 
>     7. GET /threshold/{thold_id}/results/{gen_id}
> 
>     8. GET /t-digest/{tdigest_id}/results/{gen_id}
> 
>     9. DELETE /generator/{id}
> 
>     10. DELETE /threshold/{thold_id}
> 
>     11. DELETE /t-digest/{tdigest_id}
> 
> 
> However, if you include dynamic results as parts of the config/results, you get:
> 
>     1. POST /generator
> 
>     2. POST /generator/{id}/start
> 
>     3. POST /generator/{id}/stop
> 
>     4. GET /generator-result/{result-id}
> 
>     5. DELETE /generator/{id}
> 
> 
> Using top-level entities requires over twice as many REST operations as inlining the dynamic results. Additionally, the client must also do more record keeping, because it needs to keep track of the id's for every dynamic result it has created. Making dynamic results first-class properties of the generators provides better scaling and is easier to use.
> 
> @jhiga dynamic results configurations are included in the POST to the generator's start endpoint. They aren't explicitly stored anywhere in the REST API. The results could include sufficient context to determine what they were, e.g. result property, relation, etc.

@DerangedMonkeyNinja, I remember we conclude that dynamic results lifetime not strongly coupled with generator or generator results, your opinion was changed? @jhiga what do you think about generator and dynamic results coupling?

@DerangedMonkeyNinja, in your case client will receive pack of dynamic results and should be process it additionally, includes full comparison of contexts (configurations of dynamic results), to define particular result he needs. Not user-friendly approach, in my opinion. Besides in your approach client should restarting generator each time he want change the set of dynamic results and each time pass full set of configurations of dynamic results. Even when he wants add or delete only one.

loskutnikov-spirent:
@jhiga, @DerangedMonkeyNinja. Are there any additional suggestions? Are you satisfied with the current API?

loskutnikov-spirent:
Common Dynamic Results API + swagger generated files.

PR #311, opened 2020-08-11
Fix statistics erase bug

loskutnikov-spirent:
Fix the the bug, appeared in #303 causing acceptance_test failing.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/311)


loskutnikov-spirent:
Failing of acceptance_tests was caused by incorrect using of range-based for loop with erase method of container. 
Range-based for loop replaced by regular for loop with iterator.

PR #317, opened 2020-08-18
Deleting Block File from the disk

### LINKED ISSUE #314: Delete Block File does not delete the file on disk.
Expecting the file to be deleted on disk.

**Steps.**

**1. Create a block file:**
```
POST /block-files
{
    "id": "qwe2",
    "file_size": 1024,
    "init_percent_complete": 0,
    "path": "/tmp/foo",
    "state": "init"
}
Response: 201 Created
```

**2. Delete a block file:**
```
DELETE /block-files/qwe2
Response: 204 No Content
```

**3. Get a block file.**
```
GET /block-files/qwe2
Response: 404 Not Found
```

**3.  Access the host and the file still exists at /tmp/foo**

****Note:** If you try to delete the file during initialization or after it's ready, it'll have the same outcome where it'll delete it from the api resource, but it doesn't delete it from disk.

abelov-spirent:
Signed-off-by: Artem Belov <artem.belov@spirent.com>
Fixes #314

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/317)

### DIFF
--- a/src/modules/block/file_stack.cpp
+++ b/src/modules/block/file_stack.cpp
@@ -124,6 +124,13 @@ file_stack::delete_block_file(const std::string& id)
     if (m_block_files.count(id) && m_block_files.at(id)->fd())
         return tl::make_unexpected(deletion_error_type::BUSY);
 
+    auto f = m_block_files.at(id);
+    if (remove(f->get_path().c_str()))
+        OP_LOG(OP_LOG_ERROR,
+               "Cannot delete block file %s from the disk: %s",
+               f->get_path().c_str(),
+               strerror(errno));
+
     if (m_block_files.erase(id) <= 0)
         return tl::make_unexpected(deletion_error_type::NOT_FOUND);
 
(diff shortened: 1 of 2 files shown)


abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/317)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/block/file_stack.cpp, line 141 at r1](https://reviewable.io/reviews/spirent/openperf/317#-MFBCejl3IK0WlLBcGBy:-MFUnYRH7j72gfCNDTxQ:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/bd504284277de127da0a027037c0d3f3f82ce36d/src/modules/block/file_stack.cpp#L141)):*

I'm confused by the need to keep this check here.  You *know* the id is in the collection from the `count()` call above. If you end up taking the if branch here, then your code has serious problems and should just crash. Suggest eliminating the branch and just calling `erase()`.
</blockquote></details>

Done.

PR #318, opened 2020-08-18
Fix Openperf fault on block file duplicate

### LINKED ISSUE #312: OpenPerf terminates when creating more than one block file with the same path when the path no longer exists
**Steps:**
**1. Create a block file**
```
POST /block-files
{
    "id": "qwe",
    "file_size": 1024,
    "init_percent_complete": 0,
    "path": "/tmp/foo",
    "state": "init"
}

```
**2. Get a block file**
```
GET /block-files/qwe
{
  "file_size": 1024,
  "id": "qwe",
  "init_percent_complete": 100,
  "path": "/tmp/foo",
  "state": "ready"
}
```

**3. Delete /tmp/foo from the host**

**4. Try to create a new block file with a different id, but same path:**
```
POST /block-files
{
    "id": "qwe1",
    "file_size": 1024,
    "init_percent_complete": 0,
    "path": "/tmp/foo",
    "state": "init"
}
```

**Response:**
Error: Server returned nothing (no headers, no data)

OpenPerf is now no longer reachable on the host.

abelov-spirent:
Fixes #312

Signed-off-by: Artem Belov <artem.belov@spirent.com>

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/318)

### DIFF
--- a/src/modules/block/file_stack.cpp
+++ b/src/modules/block/file_stack.cpp
@@ -87,6 +87,10 @@ file_stack::create_block_file(const model::file& block_file_model)
             return tl::make_unexpected("File with path "
                                        + block_file_model.path()
                                        + " already exists.");
+        if (code)
+            OP_LOG(OP_LOG_WARNING,
+                   "Unexpected file operation result: %s",
+                   code.message().c_str());
     }
 
     if (block_file_model.size() <= sizeof(virtual_device_header)) {
--- a/src/modules/block/file_stack.cpp
+++ b/src/modules/block/file_stack.cpp
@@ -81,8 +81,9 @@ file_stack::create_block_file(const model::file& block_file_model)
                                    + " already exists.");
 
     for (const auto& blkfile_pair : m_block_files) {
-        if (std::filesystem::equivalent(block_file_model.path(),
-                                        blkfile_pair.second->path()))
+        std::error_code code;
+        if (std::filesystem::equivalent(
+                block_file_model.path(), blkfile_pair.second->path(), code))
             return tl::make_unexpected("File with path "
                                        + block_file_model.path()
                                        + " already exists.");
--- a/src/modules/block/file_stack.cpp
+++ b/src/modules/block/file_stack.cpp
@@ -83,14 +83,15 @@ file_stack::create_block_file(const model::file& block_file_model)
     for (const auto& blkfile_pair : m_block_files) {
         std::error_code code;
         if (std::filesystem::equivalent(
-                block_file_model.path(), blkfile_pair.second->path(), code))
+                block_file_model.path(), blkfile_pair.second->path(), code)) {
             return tl::make_unexpected("File with path "
                                        + block_file_model.path()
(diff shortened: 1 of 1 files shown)


abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/318)* status: all files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/block/file_stack.cpp, line 90 at r1](https://reviewable.io/reviews/spirent/openperf/318#-MFB5ffs6oV2xYP1l9aa:-MFUnqXb7BHz-H02LYd4:b7kv8dr) ([raw file](https://github.com/spirent/openperf/blob/dc3ab906237f7dfca25f8ff910c9146c95c698e1/src/modules/block/file_stack.cpp#L90)):*
> ```
> std::filesystem::equivalent
> ```
Returns error if file not found. Do you have any other suggestion on such situation?

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/318)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/block/file_stack.cpp, line 90 at r1](https://reviewable.io/reviews/spirent/openperf/318#-MFB5ffs6oV2xYP1l9aa:-MFjUXGj5uZIfBiBVIBb:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/dc3ab906237f7dfca25f8ff910c9146c95c698e1/src/modules/block/file_stack.cpp#L90)):*

It looks like the `code` variable is unused other than in the `std::filesystem::equivalent` function...

All I see in this diff is two additions...

```
+ std::error_code code;

if (std::filesystem::equivalent(block_file_model.get_path(),
-                                       blkfile_pair.second->get_path())
+                                      blkfile_pair.second->get_path(),
+                                      code))
```
Does the code variable actually get used anywhere?

The commit message says this fixes a fault, but it's not clear to me how the change above changes behavior at all.
</blockquote></details>

Done.

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/318)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/block/file_stack.cpp, line 90 at r2](https://reviewable.io/reviews/spirent/openperf/318#-MFjzjTIEpnpH5Hs6MlU:-MFmuDI34vrVXYU2Zbcr:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/540d9ffae762542911890872f58c2933b5a7a8dc/src/modules/block/file_stack.cpp#L90)):*

I'm not sure this is better...  Is it possible for equivalent to return false AND have an error in `code`?  In that case is the false result even valid? Seems like `code` should be checked first before consulting the function result...
</blockquote></details>

Done.

PR #325, opened 2020-08-28
TVLP

abelov-spirent:
Signed-off-by: Artem Belov artem.belov@spirent.com

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/325)


abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 17 of 156 files reviewed, 8 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[mk/swagger_model.mk, line 13 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGA_6GC-xN4ANzDn1YL:-MGNCIIG6uK39m9a7gvR:b-l1haco) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/mk/swagger_model.mk#L13)):*

Why does this code need `-fPIC`?
</blockquote></details>

It is required by dynamically loaded modules, like TVLP
___
*[src/modules/tvlp/controller.cpp, line 67 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGANC1T-HJCQYwQ7SDU:-MGND2z-6OBLoHJgDc6y:bbklnbm) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/controller.cpp#L67)):*

Is it ok to call `start` on a running controller?  I would guess not, but the code doesn't prevent that from happening or complain if it does.
</blockquote></details>

It is prevented by `controller_stack.cpp`. Also added that check into `controller`
Done
___
*[src/modules/tvlp/worker.hpp, line 31 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAPdWP59vPBfZFcTRG:-MGNOeMHCdT9X3-jxCOz:bd9mb3v) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.hpp#L31)):*

This seems excessive...  What's the rationale for picking a 15 second timeout?
</blockquote></details>

I am working with docker container and sometimes process freezes for a few seconds. Changed to 5 seconds, seems to be enough
Done.
___
*[src/modules/tvlp/worker.hpp, line 33 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAZ2yE6DmrmydU7R2V:-MGano4J-mI02ZdXrg7g:b-qas30) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.hpp#L33)):*

At a high level, I'm not ecstatic about using OOP here as this is going to result in a lot of redundant code. The logic for operating generators is the same for all generators. Additionally, handling JSON and HTTP codes is also identical. I think an approach that leverages that similarity will resort in a smaller and cleaner code base. Suggest using templates and/or CRTP instead of object heirarchies.
</blockquote></details>

OOP is still required to keep it more configurable for future generators or features. Made more default realizations to get cleaner code base. Done
___
*[src/modules/tvlp/worker.cpp, line 4 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAUglW-Wsa1LzNOV3i:-MGbNTiP52IkFqK9kYle:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.cpp#L4)):*

I'm concerned that this worker appears to have little concern for its actual state.  It looks like a call to the wrong state will either block until a profile run is completed or ignored previous error. Why does this code let clients call `start()` when the worker has started? Why can clients call `start()` when there is an error? There seems to be an implicit assumptions that clients will always do "the right thing" here.
</blockquote></details>

Done.
___
*[src/modules/tvlp/worker.cpp, line 64 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGASIFe2dTRtjcFfIqh:-MGbwkxVEeXsAqzoKslx:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.cpp#L64)):*

What prevents a race on this vector?
</blockquote></details>

Done.
___
*[src/modules/tvlp/models/tvlp_config.hpp, line 62 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAWzViCF5GAy75ienX:-MGNQu3F3p1PFX0c7W8z:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/models/tvlp_config.hpp#L62)):*

There is no reason to assign a 0 length string. The default string constructor will construct an empty string.
(This comment applies globally to this review).
</blockquote></details>

Done.
___
*[src/modules/tvlp/workers/block.cpp, line 27 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAYDHF8Rjv5xRnAav4:-MGbNZ5j3s4KCuz2je9c:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/workers/block.cpp#L27)):*

I don't understand this logic.  Why does this function accept return codes that the swagger specification doesn't allow?
</blockquote></details>

Done.

DerangedMonkeyNinja (changes requested):
> Reviewed 29 of 135 files at r1, 22 of 40 files at r7.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 67 of 156 files reviewed, 8 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[src/modules/tvlp/worker.hpp, line 31 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAPdWP59vPBfZFcTRG:-MGiwkxg00xWOVlt8OzE:b-pqcrvr) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.hpp#L31)):*

I am working with docker container and sometimes process freezes for a few seconds. Changed to 5 seconds, seems to be enough
Done.
</blockquote></details>

That seems like a problem. Have you done any investigation as to why? How does that affect playback?  Since I think there are only sleeps between events, I guess the playback just becomes out of sync?
___
*[src/modules/tvlp/worker.hpp, line 33 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAZ2yE6DmrmydU7R2V:-MGj1nxJ2IHJFovXTc7A:b-pemcb6) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.hpp#L33)):*

OOP is still required to keep it more configurable for future generators or features. Made more default realizations to get cleaner code base. Done
</blockquote></details>

I disagree with your assertion. Fundamentally you're writing a bunch of boilerplate code for different types.  For example, each concrete worker object is just boilerplate with a type specific create function.  Those `send_create` functions could all be a single template.
___
*[src/modules/tvlp/worker.cpp, line 67 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGj04vg6D8Tlin6SPqh:-MGj04vg6D8Tlin6SPqi:b9ecwcl) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/src/modules/tvlp/worker.cpp#L67)):*
> ```cpp
>         auto res = m_scheduler_thread.get();
>         if (res)
>             *m_result_ptr = res.value();
> ```

Isn't this pointer updated by the schedule thread regularly?  Why would this ever be necessary?
___
*[src/modules/tvlp/worker.cpp, line 88 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGj0gnm1gCbttykMmgn:-MGj0gnm1gCbttykMmgo:b-u692p3) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/src/modules/tvlp/worker.cpp#L88)):*
> ```cpp
> 
>     const_cast<tvlp_worker_t*>(this)->fetch_future_results();
>     if (state == model::READY) return *m_result_ptr;
> ```

There is still a race here.  Since you're dereferencing the pointer, you're returning a copy of the pointer data.  You need to return the actual shared pointer since that is what is atomically updated and that is what handles reference counting for when the schedule thread's copy is destructed.
___
*[src/modules/tvlp/worker.cpp, line 94 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGj-sJ32k-2Ntf5L1X5:-MGj-sJ32k-2Ntf5L1X6:bf6crcl) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/src/modules/tvlp/worker.cpp#L94)):*
> ```cpp
>     std::shared_ptr<model::json_vector> result_copy =
>         std::make_shared<model::json_vector>();
>     auto old_ptr = std::atomic_exchange(result_ptr, result_copy);
> ```

Same comment as below.
___
*[src/modules/tvlp/worker.cpp, line 107 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGj-Ej5BlAstoUpwNNg:-MGj-Ej5BlAstoUpwNNh:b-5tv96u) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/src/modules/tvlp/worker.cpp#L107)):*
> ```cpp
>         std::make_shared<model::json_vector>(result);
>     auto old_ptr = std::atomic_exchange(&m_result_ptr, result_copy);
>     *old_ptr = result;
> ```

Why do you update the old_ptr and swap again?  Doesn't `m_result_ptr` already point to the current result at this point?

Additionally, this seems like a very inefficient approach to sharing data between two threads. Each TVLP step increases the size of this result vector and as the test progresses, this function will copy increasingly larger chunks of data that never changes. 

There are a couple of different ways to handle this... 

1) Use message passing; have the schedule thread send updated results to the controlling thread and have the controlling thread own the result aggregate.
2) Use an immutable data structure from `deps/immer` to only update the data that changes.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 67 of 156 files reviewed, 7 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[mk/swagger_model.mk, line 13 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGA_6GC-xN4ANzDn1YL:-MGj4p7J0xtViee_GSwJ:b-l74q4i) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/mk/swagger_model.mk#L13)):*

It is required by dynamically loaded modules, like TVLP
</blockquote></details>

Right, but the swagger code isn't dynamic, is it?

DerangedMonkeyNinja (changes requested):
> Reviewed 3 of 14 files at r4, 15 of 40 files at r7, 1 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 76 of 157 files reviewed, 9 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[tests/aat/common/helper/block.py, line 5 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGypbD_03BtncHlkznR:-MGypbD_03BtncHlkznS:b-ky75og) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/tests/aat/common/helper/block.py#L5)):*
> ```Python
> import time
> 
> def get_block_dynamic_results_fields():
> ```

nit: idiomatic python style uses two felines between function blocks. There is a mix of single and double spaces in all of these helper files now.
___
*[tests/aat/spec/config.yaml, line 6 at r11](https://reviewable.io/reviews/spirent/openperf/325#-MGyoPJ_8XCZM6xh_Bsw:-MGyoPJaArCszRMVv1ue:b-odkzce) ([raw file](https://github.com/spirent/openperf/blob/5d3cf3c2cd9447144c56191486ec9d927c47322c/tests/aat/spec/config.yaml#L6)):*
> ```YAML
>     # NOTE: --force-unlink will NOT verify if another copy of openperf is running.
>     # FIXME: Remove --force-unlink. This is a hacky, temporary workaround for reliability.
>     command: '../../build/openperf-linux-x86_64-testing/bin/openperf --modules.packetio.dpdk.test-mode --core.log.level debug --modules.socket.force-unlink --modules.packetio.dpdk.options="-m256m,--no-huge" --modules.packetio.dpdk.port-ids="port0=0, port1=1" -p 9000 -L "libopenperf_tvlp.so" -m "../../build/openperf-linux-x86_64-testing/plugins/" >> openperf.log 2>&1'
> ```

Can we have a dedicated spec for tvlp?  This would avoid editing the command line here and modifying other configuration files.

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 63 of 149 files reviewed, 8 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[mk/swagger_model.mk, line 13 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGA_6GC-xN4ANzDn1YL:-MGsxIQI2EH7wjXdtgPC:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/mk/swagger_model.mk#L13)):*

Right, but the swagger code isn't dynamic, is it?
</blockquote></details>

Done.
___
*[src/modules/tvlp/worker.hpp, line 31 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAPdWP59vPBfZFcTRG:-MHBOEDuAnVR8xHqyCgD:blt27pj) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.hpp#L31)):*

That seems like a problem. Have you done any investigation as to why? How does that affect playback?  Since I think there are only sleeps between events, I guess the playback just becomes out of sync?
</blockquote></details>

I found two reasons of long response:
- On stop command generator is waiting for last operation finished. It could happen on huge block size. (Does not affect on playback results)
- Pistache has two possible issues that could affect on TVLP. I found that response is added to the queue instantly but it takes few seconds to actually send it: (https://github.com/oktal/pistache/issues/49, https://github.com/oktal/pistache/issues/88)
___
*[src/modules/tvlp/worker.hpp, line 33 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAZ2yE6DmrmydU7R2V:-MHBPuFvDgFCKVAfcO_J:b-lklm46) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.hpp#L33)):*

I disagree with your assertion. Fundamentally you're writing a bunch of boilerplate code for different types.  For example, each concrete worker object is just boilerplate with a type specific create function.  Those `send_create` functions could all be a single template.
</blockquote></details>

OOP gives maximum flexibility for operations. Template will put limits. Also workers implementations contains final endpoints.
___
*[src/modules/tvlp/worker.cpp, line 67 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGj04vg6D8Tlin6SPqh:-MHBQSdp--sFIxcAGggD:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/src/modules/tvlp/worker.cpp#L67)):*

Isn't this pointer updated by the schedule thread regularly?  Why would this ever be necessary?
</blockquote></details>

Done.
___
*[src/modules/tvlp/worker.cpp, line 88 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGj0gnm1gCbttykMmgn:-MHBQVGyAxZeG0YWp39J:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/src/modules/tvlp/worker.cpp#L88)):*

There is still a race here.  Since you're dereferencing the pointer, you're returning a copy of the pointer data.  You need to return the actual shared pointer since that is what is atomically updated and that is what handles reference counting for when the schedule thread's copy is destructed.
</blockquote></details>

Done.
___
*[src/modules/tvlp/worker.cpp, line 94 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGj-sJ32k-2Ntf5L1X5:-MHBQTZN8ORILkpiXEdE:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/src/modules/tvlp/worker.cpp#L94)):*

Same comment as below.
</blockquote></details>

Done.
___
*[src/modules/tvlp/worker.cpp, line 107 at r7](https://reviewable.io/reviews/spirent/openperf/325#-MGj-Ej5BlAstoUpwNNg:-MHBQiqd0BVGb1NxcEiP:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/9193ac23294316b19ecebc40669849d6bc70d549/src/modules/tvlp/worker.cpp#L107)):*

Why do you update the old_ptr and swap again?  Doesn't `m_result_ptr` already point to the current result at this point?

Additionally, this seems like a very inefficient approach to sharing data between two threads. Each TVLP step increases the size of this result vector and as the test progresses, this function will copy increasingly larger chunks of data that never changes. 

There are a couple of different ways to handle this... 

1) Use message passing; have the schedule thread send updated results to the controlling thread and have the controlling thread own the result aggregate.
2) Use an immutable data structure from `deps/immer` to only update the data that changes.
</blockquote></details>

Done.
___
*[tests/aat/spec/config.yaml, line 6 at r11](https://reviewable.io/reviews/spirent/openperf/325#-MGyoPJ_8XCZM6xh_Bsw:-MHBWQKd-MaZGyoyHqw0:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/5d3cf3c2cd9447144c56191486ec9d927c47322c/tests/aat/spec/config.yaml#L6)):*

Can we have a dedicated spec for tvlp?  This would avoid editing the command line here and modifying other configuration files.
</blockquote></details>

Done.

DerangedMonkeyNinja (changes requested):
> Reviewed 5 of 6 files at r8, 3 of 4 files at r9, 2 of 2 files at r10, 1 of 1 files at r11, 12 of 19 files at r12.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 71 of 149 files reviewed, 3 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/325#-MHGnzad4WPd_-NYFr3k:-MHGnzae7araNyw3N20i:b-axm24d) (no related file):*
Is there a linking/compilation reason for moving all those functions to `src/swagger/converters`?

---
___
*[src/modules/tvlp/worker.hpp, line 31 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAPdWP59vPBfZFcTRG:-MHGkNq-2oORDpvCzpUf:b5zvemq) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.hpp#L31)):*

I found two reasons of long response:
- On stop command generator is waiting for last operation finished. It could happen on huge block size. (Does not affect on playback results)
- Pistache has two possible issues that could affect on TVLP. I found that response is added to the queue instantly but it takes few seconds to actually send it: (https://github.com/oktal/pistache/issues/49, https://github.com/oktal/pistache/issues/88)
</blockquote></details>

Are version of pistache is rather old, would an update bring in useful changes?
___
*[src/modules/tvlp/worker.cpp, line 75 at r12](https://reviewable.io/reviews/spirent/openperf/325#-MHGnRMa5DkqOFjvJHTd:-MHGnRMa5DkqOFjvJHTe:b1go96b) ([raw file](https://github.com/spirent/openperf/blob/8a256e4a3dad8b6e25ac5c4d2d713fe64972e193/src/modules/tvlp/worker.cpp#L75)):*
> ```cpp
> duration tvlp_worker_t::offset() const { return m_state.offset.load(); }
> 
> model::json_vector tvlp_worker_t::results() const
> ```

Will this vector be used outside of the worker's lifetime?

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 71 of 149 files reviewed, 3 unresolved discussions (waiting on @DerangedMonkeyNinja)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/325#-MHGnzad4WPd_-NYFr3k:-MHGpTxT7g3z7hBbKoRS:b-5h4nf2) (no related file):*

Is there a linking/compilation reason for moving all those functions to `src/swagger/converters`?
</blockquote></details>

Yes. Tvlp should have access to this functions even if module will not be linked in the future

---
___
*[src/modules/tvlp/worker.hpp, line 31 at r6](https://reviewable.io/reviews/spirent/openperf/325#-MGAPdWP59vPBfZFcTRG:-MHGpfI94JCUHYv48fvl:bd0phl) ([raw file](https://github.com/spirent/openperf/blob/eec328c3c485325bef04a7cf01f73becdd3ac9c6/src/modules/tvlp/worker.hpp#L31)):*

Are version of pistache is rather old, would an update bring in useful changes?
</blockquote></details>

I already checked that. Did not help.
___
*[src/modules/tvlp/worker.cpp, line 75 at r12](https://reviewable.io/reviews/spirent/openperf/325#-MHGnRMa5DkqOFjvJHTd:-MHGpiq1DDbSvFZ8iXvi:b-7xt2jw) ([raw file](https://github.com/spirent/openperf/blob/8a256e4a3dad8b6e25ac5c4d2d713fe64972e193/src/modules/tvlp/worker.cpp#L75)):*

Will this vector be used outside of the worker's lifetime?
</blockquote></details>

Yes.

DerangedMonkeyNinja (changes requested):
> Reviewed 16 of 135 files at r1, 1 of 40 files at r7, 17 of 18 files at r13.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 87 of 149 files reviewed, 7 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[src/modules/tvlp/controller.hpp, line 14 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwQNxF8PNXQgX7eiBy:-MHwQNxGAARhqWQhn6j3:bum13gm) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/controller.hpp#L14)):*
> ```cpp
> namespace openperf::tvlp::internal {
> 
> using framework_controller = openperf::framework::generator::controller;
> ```

I think the general consensus is to avoid renaming things via `using` in header files, since everything that includes this file will get the using declarations. Better to use them in the class or in the compilation unit instead.
___
*[src/modules/tvlp/controller.cpp, line 148 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwR9Q471LsW2gJESHa:-MHwR9Q471LsW2gJESHb:b-tpeqmm) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/controller.cpp#L148)):*
> ```cpp
>     };
> 
>     m_error = "";
> ```

Prefer `clear()` over re-assignment here.
___
*[src/modules/tvlp/controller_stack.hpp, line 13 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwRMJH8rd6arUkgssf:-MHwRMJH8rd6arUkgssg:blqkrof) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/controller_stack.hpp#L13)):*
> ```cpp
> namespace openperf::tvlp::internal {
> 
> using tvlp_controller_ptr = std::shared_ptr<controller_t>;
> ```

Same comment as earlier.  Are clients using these types directly?
___
*[src/modules/tvlp/workers/memory.cpp, line 10 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwWq8831bc48N2B4MQ:-MHwWq89A3iJkYoKmj9d:b-x6aggx) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/workers/memory.cpp#L10)):*
> ```cpp
> namespace openperf::memory::api {
> std::pair<Pistache::Http::Code, std::optional<std::string>>
> to_error(const reply::error& error);
> ```

Is there no header for this function?  I'd prefer you include a header for this instead of (re)defining it here. Since this file is going to get dynamically linked, I think any change to this functions signature will result in a run-time error for an undefined symbol.
___
*[src/modules/tvlp/workers/packet.cpp, line 10 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwXF8q4k5sFgQGImuL:-MHwXF8q4k5sFgQGImuM:bxyo3gv) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/workers/packet.cpp#L10)):*
> ```cpp
> 
> namespace openperf::packet::generator::api {
> const char* to_string(const reply_error& error);
> ```

Same comment as earlier.  Please use a header for function declarations.
___
*[src/swagger/converters/packetio.hpp, line 7 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwYQHHDtLzMc-n9-bt:-MHwYQHHDtLzMc-n9-bu:b-y316xd) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/swagger/converters/packetio.hpp#L7)):*
> ```cpp
> 
> namespace openperf::packetio {
> static const std::string empty_id_string = "none";
> ```

What is this for?  Empty strings are pretty easy to check, so I don't think they should have a sentinel value.
___
*[src/swagger/converters/tvlp.cpp, line 115 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwYsbb6H1pvuubeC8H:-MHwYsbb6H1pvuubeC8I:bth5fwz) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/swagger/converters/tvlp.cpp#L115)):*
> ```cpp
>         if (!val["memory"].is_null()) {
>             auto newItem = std::make_shared<TvlpProfile_memory>();
>             from_json(val.at("memory"), *newItem);
> ```

I guess you're using `from_json` here to avoid the copy of `auto x = std::make_shared<T>(val.get<T>())`?

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325)* status: 78 of 149 files reviewed, 7 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/tvlp/controller.hpp, line 14 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwQNxF8PNXQgX7eiBy:-MHyKGkBDCE0BSsB8mpJ:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/controller.hpp#L14)):*

I think the general consensus is to avoid renaming things via `using` in header files, since everything that includes this file will get the using declarations. Better to use them in the class or in the compilation unit instead.
</blockquote></details>

Done.
___
*[src/modules/tvlp/controller.cpp, line 148 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwR9Q471LsW2gJESHa:-MHyKOtC8nAYuo9LCjSr:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/controller.cpp#L148)):*

Prefer `clear()` over re-assignment here.
</blockquote></details>

Done.
___
*[src/modules/tvlp/controller_stack.hpp, line 13 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwRMJH8rd6arUkgssf:-MHyKTawEy24KQWf1rF5:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/controller_stack.hpp#L13)):*

Same comment as earlier.  Are clients using these types directly?
</blockquote></details>

Done.
___
*[src/modules/tvlp/workers/memory.cpp, line 10 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwWq8831bc48N2B4MQ:-MHyKlQe3PvSa-Bu_Aa1:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/workers/memory.cpp#L10)):*

Is there no header for this function?  I'd prefer you include a header for this instead of (re)defining it here. Since this file is going to get dynamically linked, I think any change to this functions signature will result in a run-time error for an undefined symbol.
</blockquote></details>

Done.
___
*[src/modules/tvlp/workers/packet.cpp, line 10 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwXF8q4k5sFgQGImuL:-MHyM7ynBlUAf6xlAQ6k:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/modules/tvlp/workers/packet.cpp#L10)):*

Same comment as earlier.  Please use a header for function declarations.
</blockquote></details>

Done.
___
*[src/swagger/converters/packetio.hpp, line 7 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwYQHHDtLzMc-n9-bt:-MHyMDB-1r_62hZne08l:bpnowls) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/swagger/converters/packetio.hpp#L7)):*

What is this for?  Empty strings are pretty easy to check, so I don't think they should have a sentinel value.
</blockquote></details>

I have moved that from packetio module as is
___
*[src/swagger/converters/tvlp.cpp, line 115 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwYsbb6H1pvuubeC8H:-MHyNaW3-vvXEImpMEca:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/swagger/converters/tvlp.cpp#L115)):*

I guess you're using `from_json` here to avoid the copy of `auto x = std::make_shared<T>(val.get<T>())`?
</blockquote></details>

Done.

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 8 of 9 files at r14.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/325#-:-MI1-CifAI07V_7u0sQU:bnfp4nl)* status: 86 of 149 files reviewed, all discussions resolved (waiting on @DerangedMonkeyNinja)

___
*[src/swagger/converters/tvlp.cpp, line 115 at r13](https://reviewable.io/reviews/spirent/openperf/325#-MHwYsbb6H1pvuubeC8H:-MI0zvtB6tZ753Jgm1ak:b-69ub3j) ([raw file](https://github.com/spirent/openperf/blob/b5afbdc6162997ed4926a99e1124677e465cea21/src/swagger/converters/tvlp.cpp#L115)):*

Done.
</blockquote></details>

That really was just a question, but ok. :)

PR #327, opened 2020-08-31
CPU Generator SIMD

### LINKED ISSUE #310: Use ISPC to create SIMD kernels for CPU load generation
For example, SSE, AVX, AVX2, etc.

loskutnikov-spirent:
Closes #310

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/327)


loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/327)* status: 15 of 21 files reviewed, 9 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[mk/cpu.mk, line 72 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4rSvB-mSF2IkH8lcL:-MG5t_N43qrShTSaxAqa:bss4a2b) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/mk/cpu.mk#L72)):*

This options was in the PGA defines because this instruction is required for CRC calculations.  I don't think you need it for the CPU module, do you?
</blockquote></details>

Removed.
___
*[src/modules/cpu/function_wrapper.hpp, line 56 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4sO0oERRc48t4eAc9:-MG5zcNPCvhYSJ3S7Z3L:b-s9dczu) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/src/modules/cpu/function_wrapper.hpp#L56)):*

Suggest using `openperf::cpu::simd` as the namespace here so that these functions are clearly part of the cpu module.
</blockquote></details>

The namespace here should corresponds to namespace at `ISPC_FUNCTION_WRAPPER_INIT`. `ISPC_FUNCTION_WRAPPER_INIT` is called inside `openperf::cpu::internal` namespace, hence result namespace will be `openperf::cpu::internal::simd`.
___
*[src/modules/cpu/generator.hpp, line 45 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4sewyBxdxonlsZ7MF:-MG5tkCv11YZVTh6ciiS:b-2s3cla) ([raw file](https://github.com/spirent/openperf/blob/d1579b9c560a206337ad735db98662241cea6cfc/src/modules/cpu/generator.hpp#L45)):*

Suggest making this a free function as there is nothing specific to the generator that this function needs access to.  Additionally, wouldn't you need to make use of this function in the REST configuration validation anyway?
</blockquote></details>

The method `is_supported` is no longer exists at all. Was replaced by `enabled()` and `available()` functions.
___
*[src/modules/cpu/instruction_set.hpp, line 36 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4t7WsCcrNF7vZqzFC:-MG5uWWoCZ6EftPtaxcV:b-jdskyv) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/src/modules/cpu/instruction_set.hpp#L36)):*

Why is `auto` here?  It doesn't look like the REST API allows an automatic mode to be configured.
</blockquote></details>

I added `AUTO` because ISPC generates function without suffix. I think that the internal part may not corresponds to REST interface literally. Of course I can add `AUTO` to API specification, if you want.
___
*[src/modules/cpu/matrix.hpp, line 10 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4tWai2eTxEmIQJIiS:-MG5vbvd7ZkVIIySdCoy:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/src/modules/cpu/matrix.hpp#L10)):*

prefer `type[]` over `type*` for array arguments. It's more explicit for human readers even if the compiler treats it as the same value.
</blockquote></details>

Done.
___
*[src/modules/cpu/target_ispc.hpp, line 47 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4wqqP4XzO4T983p0f:-MG5vcZ5ATxRGs1NvC0C:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/src/modules/cpu/target_ispc.hpp#L47)):*

You can avoid performing this `const_cast` if you update the function signature by marking matrix a and b const arguments (they should be) and adding the `mutable` tag to matrix r.
</blockquote></details>

Done.
___
*[src/modules/cpu/ispc/matrix.ispc, line 3 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4wenc5J2Pz5ikE2sc:-MG5vdfh7U5SZr8j-Qfz:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/src/modules/cpu/ispc/matrix.ispc#L3)):*

matrix a and b should be marked const
</blockquote></details>

Done.
___
*[src/modules/cpu/scalar/matrix.cpp, line 8 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4wis_9C2nWZijvLjl:-MG5veDJ5UEzAbYRKQxV:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/src/modules/cpu/scalar/matrix.cpp#L8)):*

Matrix a and b should be marked const.
</blockquote></details>

Done.
___
*[tests/unit/modules/cpu/test_ispc_target.cpp, line 101 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4vqgz1POaz6RNfPem:-MG5wC5d8l_oxVqxZrie:bl0d04m) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/tests/unit/modules/cpu/test_ispc_target.cpp#L101)):*

Two comments:
1) The scalar instruction set doesn't need to be run twice.
2) The test should require that at least one vector instruction set is run and matches the scalar reference result.
</blockquote></details>

1. Scalar function runs twice, because the first run gives reference matrix for testing other instruction sets correctness. The second run needs for checking `enabled()` and `available()` methods and running inside `BENCHMARK` section.
2. The unit test already compares result of an instruction set with scalar reference result.

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/327)* status: 11 of 21 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/cpu/instruction_set.hpp, line 36 at r1](https://reviewable.io/reviews/spirent/openperf/327#-MG4t7WsCcrNF7vZqzFC:-MGISI3gBQgHU20spMFu:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/4aab4e63ed4c3c1a9394ad806ce7618796b8c4fb/src/modules/cpu/instruction_set.hpp#L36)):*

No, the version without any suffix should never be called. It's a convenience function to redirect to the most advanced instruction set supported by the CPU. Since the user is explicitly telling us which instruction set to run, calling the non-suffixed version could do the wrong thing. Hence, the code shouldn't support handling the auto case at all.
</blockquote></details>

Done.

DerangedMonkeyNinja (changes requested):
> Reviewed 10 of 10 files at r3.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/327)* status: all files reviewed, 1 unresolved discussion (waiting on @loskutnikov-spirent)

___
*[tests/unit/modules/cpu/test_ispc_target.cpp, line 5 at r3](https://reviewable.io/reviews/spirent/openperf/327#-MGKftsK4jXawGuraq5U:-MGKftsK4jXawGuraq5V:brnwilg) ([raw file](https://github.com/spirent/openperf/blob/33516441c423a621749741c48862e0f7f4a3c6ad/tests/unit/modules/cpu/test_ispc_target.cpp#L5)):*
> ```cpp
> #include <algorithm>
> 
> #define ISPC_TARGET_SSE2
> ```

Shouldn't these come from the makefiles directly? Otherwise the source and the build could become out of sync?

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/327)* status: 19 of 21 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[tests/unit/modules/cpu/test_ispc_target.cpp, line 5 at r3](https://reviewable.io/reviews/spirent/openperf/327#-MGKftsK4jXawGuraq5U:-MGMNpps5-19aaZYr2e7:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/33516441c423a621749741c48862e0f7f4a3c6ad/tests/unit/modules/cpu/test_ispc_target.cpp#L5)):*

Shouldn't these come from the makefiles directly? Otherwise the source and the build could become out of sync?
</blockquote></details>

Done.

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/327#-:-MGPXctsC6ljSDVRQqBs:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #334, opened 2020-09-09
Eliminate data wrapping to a pointer in generators API

### LINKED ISSUE #321: Eliminate data wrapping to a pointer in generators API.
According to [this](https://github.com/Spirent/openperf/pull/301#pullrequestreview-462727794) comment wrapping data to pointer is no longer needed.

loskutnikov-spirent:
Closes #321

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/334)

### DIFF
--- a/src/modules/block/api.hpp
+++ b/src/modules/block/api.hpp
@@ -28,7 +28,6 @@ using device_ptr = std::unique_ptr<model::device>;
 using file_ptr = std::unique_ptr<model::file>;
 using generator_ptr = std::unique_ptr<model::block_generator>;
 using generator_result_ptr = std::unique_ptr<model::block_generator_result>;
-using string_ptr = std::unique_ptr<std::string>;
 
 static constexpr size_t err_max_length = 256;
 enum class error_type : uint8_t {
@@ -85,7 +84,7 @@ struct request_block_file_bulk_add
 
 struct request_block_file_bulk_del
 {
-    std::vector<string_ptr> ids;
+    std::vector<std::string> ids;
 };
 
 struct request_block_generator_list
@@ -113,18 +112,13 @@ struct request_block_generator_bulk_add
 
 struct request_block_generator_bulk_del
 {
-    std::vector<string_ptr> ids;
+    std::vector<std::string> ids;
 };
 
 struct request_block_generator_start
 {
-    struct start_data
-    {
-        std::string id;
-        dynamic::configuration dynamic_results;
-    };
-
-    std::unique_ptr<start_data> data;
+    std::string id;
+    dynamic::configuration dynamic_results;
 };
 
 struct request_block_generator_stop
@@ -134,18 +128,13 @@ struct request_block_generator_stop
 
 struct request_block_generator_bulk_start
 {
-    struct start_data
-    {
-        std::vector<std::string> ids;
-        dynamic::configuration dynamic_results;
-    };
-
-    std::unique_ptr<start_data> data;
+    std::vector<std::string> ids;
+    dynamic::configuration dynamic_results;
 };
 
 struct request_block_generator_bulk_stop
 {
-    std::vector<string_ptr> ids;
+    std::vector<std::string> ids;
 };
 
 struct request_block_generator_result_list
--- a/src/modules/block/api_converters.cpp
+++ b/src/modules/block/api_converters.cpp
@@ -350,10 +350,9 @@ request_block_file_bulk_add from_swagger(BulkCreateBlockFilesRequest& p_request)
 
 request_block_file_bulk_del from_swagger(BulkDeleteBlockFilesRequest& p_request)
 {
-    request_block_file_bulk_del request;
-    for (auto& id : p_request.getIds())
-        request.ids.emplace_back(std::make_unique<std::string>(id));
-    return request;
+    return request_block_file_bulk_del{
+        .ids = std::move(p_request.getIds()),
+    };
 }
 
 request_block_generator_bulk_add
@@ -369,20 +368,19 @@ from_swagger(BulkCreateBlockGeneratorsRequest& p_request)
 request_block_generator_bulk_del
 from_swagger(BulkDeleteBlockGeneratorsRequest& p_request)
 {
-    request_block_generator_bulk_del request;
-    for (auto& id : p_request.getIds())
-        request.ids.emplace_back(std::make_unique<std::string>(id));
-    return request;
+    return request_block_generator_bulk_del{
+        .ids = std::move(p_request.getIds()),
+    };
 }
 
 request_block_generator_bulk_start
 from_swagger(BulkStartBlockGeneratorsRequest& p_request)
 {
     request_block_generator_bulk_start request;
-    for (auto& id : p_request.getIds()) request.data->ids.emplace_back(id);
+    for (auto& id : p_request.getIds()) request.ids.emplace_back(id);
 
     if (p_request.dynamicResultsIsSet())
-        request.data->dynamic_results =
+        request.dynamic_results =
             dynamic::from_swagger(*p_request.getDynamicResults());
 
     return request;
@@ -391,10 +389,9 @@ from_swagger(BulkStartBlockGeneratorsRequest& p_request)
 request_block_generator_bulk_stop
 from_swagger(BulkStopBlockGeneratorsRequest& p_request)
 {
-    request_block_generator_bulk_stop request;
-    for (auto& id : p_request.getIds())
-        request.ids.emplace_back(std::make_unique<std::string>(id));
-    return request;
+    return request_block_generator_bulk_stop{
+        .ids = std::move(p_request.getIds()),
+    };
 }
 
 } // namespace openperf::block::api
--- a/src/modules/block/api_transmogrify.cpp
+++ b/src/modules/block/api_transmogrify.cpp
@@ -15,86 +15,80 @@ message::serialized_message serialize_request(request_msg&& msg)
         (message::push(serialized, msg.index())
          || std::visit(
              utils::overloaded_visitor(
-                 [&](const request_block_device_list&) -> bool {
-                     return false;
-                 },
-                 [&](const request_block_device& blkdevice) -> bool {
+                 [&](const request_block_device_list&) { return 0; },
+                 [&](const request_block_device& blkdevice) {
                      return message::push(serialized, blkdevice.id);
                  },
-                 [&](const request_block_device_init& blkdevice) -> bool {
+                 [&](const request_block_device_init& blkdevice) {
                      return message::push(serialized, blkdevice.id);
                  },
-                 [&](const request_block_file_list&) -> bool { return false; },
-                 [&](const request_block_file& blkfile) -> bool {
+                 [&](const request_block_file_list&) { return 0; },
+                 [&](const request_block_file& blkfile) {
                      return message::push(serialized, blkfile.id);
                  },
-                 [&](request_block_file_add& blkfile) -> bool {
+                 [&](request_block_file_add& blkfile) {
                      return message::push(serialized,
                                           std::move(blkfile.source));
                  },
-                 [&](const request_block_file_del& blkfile) -> bool {
+                 [&](const request_block_file_del& blkfile) {
                      return message::push(serialized, blkfile.id);
                  },
-                 [&](request_block_file_bulk_add& request) -> bool {
+                 [&](request_block_file_bulk_add& request) {
                      return message::push(serialized, request.files);
                  },
-                 [&](request_block_file_bulk_del& request) -> bool {
+                 [&](request_block_file_bulk_del& request) {
                      return message::push(
                          serialized,
                          std::make_unique<request_block_file_bulk_del>(
                              std::move(request)));
                  },
-                 [&](const request_block_generator_list&) -> bool {
-                     return false;
-                 },
-                 [&](const request_block_generator& blkgenerator) -> bool {
+                 [&](const request_block_generator_list&) { return 0; },
+                 [&](const request_block_generator& blkgenerator) {
                      return message::push(serialized, blkgenerator.id);
                  },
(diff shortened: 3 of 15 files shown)


DerangedMonkeyNinja (changes requested):
> Reviewed 8 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/334)* status: 8 of 15 files reviewed, 5 unresolved discussions (waiting on @DerangedMonkeyNinja and @loskutnikov-spirent)

___
*[src/modules/block/api_transmogrify.cpp, line 18 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtrlaz7T-Ib_SSihAj:-MGtrlaz7T-Ib_SSihAk:b4jn9id) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L18)):*
> ```cpp
>          || std::visit(
>              utils::overloaded_visitor(
>                  [&](const request_block_device_list&) -> bool {
> ```

`message::push` doesn't return a `bool`; it returns an `int`.  Additionally, specifically adding bool here is confusing, since replying as 'false' could indicate that something failed. This is a global comment.
___
*[src/modules/block/api_transmogrify.cpp, line 44 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtuKE_4ajcTJsVOO5K:-MGtuKE_4ajcTJsVOO5L:bueiedj) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L44)):*
> ```cpp
>                      return message::push(
>                          serialized,
>                          std::make_unique<request_block_file_bulk_del>(
> ```

This doesn't seem correct... What keeps the object around when that unique_ptr goes out of scope?
___
*[src/modules/block/api_transmogrify.cpp, line 169 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtss1e0yZW__n-1FUN:-MGtss1e0yZW__n-1FUO:b-mh78f8) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L169)):*
> ```cpp
>     }
>     case utils::variant_index<request_msg, request_block_file_bulk_del>(): {
>         return *std::unique_ptr<request_block_file_bulk_del>(
> ```

huh?  Why are you creating and then dereferencing a unique ptr?  Just instantiate and return the object.
___
*[src/modules/cpu/handler.cpp, line 10 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGttIVKEYc7QQN4V_BF:-MGttIVKEYc7QQN4V_BG:bw7fscq) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/cpu/handler.cpp#L10)):*
> ```cpp
> #include "modules/api/api_route_handler.hpp"
> 
> namespace opneperf::cpu {
> ```

This namespace is misspelled.
___
*[src/modules/cpu/handler.cpp, line 333 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGttBae3dNiqBpCDrqz:-MGttBaf8_6KDaq4S_jP:bpuzxip) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/cpu/handler.cpp#L333)):*
> ```cpp
>         response.send(Http::Code::Internal_Server_Error);
>     }
> } // namespace opneperf::cpu
> ```

I only noticed because this appears to have inadvertently added?

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/334)* status: 7 of 15 files reviewed, 5 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/block/api_transmogrify.cpp, line 18 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtrlaz7T-Ib_SSihAj:-MGv-2pJ0KkuSOWAedwg:bjm7pxs) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L18)):*

`message::push` doesn't return a `bool`; it returns an `int`.  Additionally, specifically adding bool here is confusing, since replying as 'false' could indicate that something failed. This is a global comment.
</blockquote></details>

Well, `int` value is uses as `bool` in this function later, thus there is no big difference in this case. Using the `bool` allows use `||` operator like this:
```
return message::push(serialized, request.id)
    || message::push(
        serialized,
        std::make_unique<dynamic::configuration>(
            std::move(request.dynamic_results)));
```
that does code more laconic and clarify.
___
*[src/modules/block/api_transmogrify.cpp, line 44 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtuKE_4ajcTJsVOO5K:-MGuyKIh7mHbdONypp25:b-ye1wpn) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L44)):*

This doesn't seem correct... What keeps the object around when that unique_ptr goes out of scope?
</blockquote></details>

This code is correct. Result of `std::make_unique` is a rvalue expression, thus will be called move constructor of `std::unique_ptr` during pass argument to `message::push`. In its turn `message::push` uses `.release()` method of `std::unique_ptr` as you know.
___
*[src/modules/block/api_transmogrify.cpp, line 169 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtss1e0yZW__n-1FUN:-MGuwe_gAJP72k_C4D9D:b-roeuq1) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L169)):*

huh?  Why are you creating and then dereferencing a unique ptr?  Just instantiate and return the object.
</blockquote></details>

Function returns by-value, dereference cause copy the object. We receive a pointer from ZMQ, and object in memory should be deleted. The `std::unique_ptr` helps avoid explicit `delete` call for object from ZMQ.
___
*[src/modules/cpu/handler.cpp, line 10 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGttIVKEYc7QQN4V_BF:-MGuxXlpAWZlBnaYKiNL:b-igt2ya) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/cpu/handler.cpp#L10)):*

This namespace is misspelled.
</blockquote></details>

Thanks for this! Renamed.
___
*[src/modules/cpu/handler.cpp, line 333 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGttBae3dNiqBpCDrqz:-MGuxqyO3-qMVxl_C-kO:b-gaify6) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/cpu/handler.cpp#L333)):*

I only noticed because this appears to have inadvertently added?
</blockquote></details>

Yes. Removed.

DerangedMonkeyNinja (changes requested):
> Reviewed 7 of 15 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/334)* status: all files reviewed, 2 unresolved discussions (waiting on @loskutnikov-spirent)

___
*[src/modules/block/api_transmogrify.cpp, line 18 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtrlaz7T-Ib_SSihAj:-MGymt-v7hg0i-RgKgz_:bowkwux) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L18)):*

Well, `int` value is uses as `bool` in this function later, thus there is no big difference in this case. Using the `bool` allows use `||` operator like this:
```
return message::push(serialized, request.id)
    || message::push(
        serialized,
        std::make_unique<dynamic::configuration>(
            std::move(request.dynamic_results)));
```
that does code more laconic and clarify.
</blockquote></details>

bool and int are implicitly convertible, so most of these bools are unnecessary. Where (push || push) is used, you can specify the return type as int and avoid the confusion where `false` means success.
___
*[src/modules/block/api_transmogrify.cpp, line 44 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtuKE_4ajcTJsVOO5K:-MGylJ4m1yig3mlIS2DD:b-olu081) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L44)):*

This code is correct. Result of `std::make_unique` is a rvalue expression, thus will be called move constructor of `std::unique_ptr` during pass argument to `message::push`. In its turn `message::push` uses `.release()` method of `std::unique_ptr` as you know.
</blockquote></details>

Ok... But message::push has an overload for `std::vector<std::unique_ptr<T>>`, so it can encode request.ids directly without having to make another copy of the request data for that `unique_ptr`.
___
*[src/modules/block/api_transmogrify.cpp, line 169 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtss1e0yZW__n-1FUN:-MGylnFp1mEMkQKPxb0G:b-4albe3) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L169)):*

Function returns by-value, dereference cause copy the object. We receive a pointer from ZMQ, and object in memory should be deleted. The `std::unique_ptr` helps avoid explicit `delete` call for object from ZMQ.
</blockquote></details>

Again, there are explicit functions for this case.  Use

```
auto reply = {.ids = message::pop_unique_vector<std::string>(msg)};
return reply;
```

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/334)* status: 12 of 15 files reviewed, 2 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/block/api_transmogrify.cpp, line 18 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtrlaz7T-Ib_SSihAj:-MH9M4rB9u9dbj3iSZac:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L18)):*

bool and int are implicitly convertible, so most of these bools are unnecessary. Where (push || push) is used, you can specify the return type as int and avoid the confusion where `false` means success.
</blockquote></details>

Done.
___
*[src/modules/block/api_transmogrify.cpp, line 44 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtuKE_4ajcTJsVOO5K:-MH9MDz5Cp532CCGRDTx:bhx8rg3) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L44)):*

Ok... But message::push has an overload for `std::vector<std::unique_ptr<T>>`, so it can encode request.ids directly without having to make another copy of the request data for that `unique_ptr`.
</blockquote></details>

I use `std::vector<std::string>` instead of `std::vector<std::unique_ptr<std::string>>`.  First variant eliminates excess wrapping each string into `std::unique_ptr`. Hence code stands more obvious and more native.
___
*[src/modules/block/api_transmogrify.cpp, line 169 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtss1e0yZW__n-1FUN:-MH9NxOFDqV2lzwHnwmr:bmvr6z9) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L169)):*

Again, there are explicit functions for this case.  Use

```
auto reply = {.ids = message::pop_unique_vector<std::string>(msg)};
return reply;
```
</blockquote></details>

There is no need pack `std::string` to `std::unique_ptr` in containers, because `std::string` has copy and move constructors, so we can operate of container in common manner.

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/334#-:-MHC03t23A-7nXhW6N6x:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/modules/block/api_transmogrify.cpp, line 44 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtuKE_4ajcTJsVOO5K:-MHByWCQ7xLDYuMSoU81:b-245fbw) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L44)):*

I use `std::vector<std::string>` instead of `std::vector<std::unique_ptr<std::string>>`.  First variant eliminates excess wrapping each string into `std::unique_ptr`. Hence code stands more obvious and more native.
</blockquote></details>

Agree that a container of strings is more idiomatic. The pointer is there to to allow moving the strings between processes instead of copying them.
___
*[src/modules/block/api_transmogrify.cpp, line 169 at r1](https://reviewable.io/reviews/spirent/openperf/334#-MGtss1e0yZW__n-1FUN:-MHBxSkHDlAtmCdX-w53:b32wtd) ([raw file](https://github.com/spirent/openperf/blob/66ee2b321cd628fe238d85d823cb1bf77110ff86/src/modules/block/api_transmogrify.cpp#L169)):*

There is no need pack `std::string` to `std::unique_ptr` in containers, because `std::string` has copy and move constructors, so we can operate of container in common manner.
</blockquote></details>

ok

PR #387, opened 2020-11-14
Add interface lookup to packetio internal api

dmorton-spirent:
* packetio clients can query for interfaces by interface ID.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/387)

### DIFF
--- a/src/modules/packetio/generic_workers.hpp
+++ b/src/modules/packetio/generic_workers.hpp
@@ -116,6 +116,12 @@ public:
         m_self->del_source(dst_id, source);
     }
 
+    tl::expected<interface::generic_interface, int>
+    interface(std::string_view interface_id) const
+    {
+        return (m_self->interface(interface_id));
+    }
+
     tl::expected<void, int> swap_source(std::string_view dst_id,
                                         packet::generic_source outgoing,
                                         packet::generic_source incoming)
@@ -181,6 +187,8 @@ private:
                       interface::generic_interface interface) = 0;
         virtual void del_interface(std::string_view port_id,
                                    interface::generic_interface interface) = 0;
+        virtual tl::expected<interface::generic_interface, int>
+        interface(std::string_view interface_id) const = 0;
         virtual tl::expected<void, int>
         add_sink(packet::traffic_direction direction,
                  std::string_view src_id,
@@ -237,6 +245,12 @@ private:
                            interface::generic_interface interface) override
         {
             m_workers.del_interface(port_id, interface);
+        }
+
+        tl::expected<interface::generic_interface, int>
+        interface(std::string_view interface_id) const override
+        {
+            return (m_workers.interface(interface_id));
         }
 
         tl::expected<void, int> add_sink(packet::traffic_direction direction,
--- a/src/modules/packetio/internal_api.hpp
+++ b/src/modules/packetio/internal_api.hpp
@@ -36,6 +36,11 @@ struct request_interface_add
     interface_data data;
 };
 
+struct request_interface
+{
+    std::string interface_id;
+};
+
 struct request_interface_del
 {
     interface_data data;
@@ -118,6 +123,11 @@ struct request_worker_ids
     packet::traffic_direction direction;
 };
 
+struct reply_interface
+{
+    interface_data data;
+};
+
 struct reply_task_add
 {
     std::string task_id;
@@ -151,7 +161,8 @@ struct reply_error
     int value;
 };
 
-using request_msg = std::variant<request_interface_add,
+using request_msg = std::variant<request_interface,
+                                 request_interface_add,
                                  request_interface_del,
                                  request_port_index,
                                  request_sink_add,
@@ -164,7 +175,8 @@ using request_msg = std::variant<request_interface_add,
                                  request_transmit_function,
                                  request_worker_ids>;
 
-using reply_msg = std::variant<reply_port_index,
+using reply_msg = std::variant<reply_interface,
+                               reply_port_index,
                                reply_task_add,
                                reply_transmit_function,
                                reply_worker_ids,
--- a/src/modules/packetio/internal_client.cpp
+++ b/src/modules/packetio/internal_client.cpp
@@ -159,6 +159,31 @@ client::del_interface(std::string_view port_id,
     return (tl::make_unexpected(EBADMSG));
 }
 
+tl::expected<interface::generic_interface, int>
+client::interface(std::string_view interface_id)
+{
+    if (interface_id.length() > name_length_max) {
+        OP_LOG(OP_LOG_ERROR,
+               "Interface ID, %.*s, is too big\n",
+               static_cast<int>(interface_id.length()),
+               interface_id.data());
+        return (tl::make_unexpected(ENOMEM));
+    }
+
+    auto request = request_interface{.interface_id = std::string(interface_id)};
+
+    auto reply = do_request(m_socket.get(), request);
+    if (!reply) { return (tl::make_unexpected(reply.error())); }
+
+    if (auto maybe_interface = std::get_if<reply_interface>(&reply.value())) {
+        return (maybe_interface->data.interface);
+    } else if (auto error = std::get_if<reply_error>(&reply.value())) {
+        return (tl::make_unexpected(error->value));
+    }
+
+    return (tl::make_unexpected(EBADMSG));
+}
+
 tl::expected<void, int> client::add_sink(packet::traffic_direction direction,
                                          std::string_view src_id,
                                          packet::generic_sink sink)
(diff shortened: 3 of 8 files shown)


dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/387)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/framework/message/serialized_message.hpp, line 249 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MM3ATO-Bzj-IU7Ww7kh:-MM3ATO-Bzj-IU7Ww7ki:bjyf8dm) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/framework/message/serialized_message.hpp#L249)):*
> ```cpp
>     pop_front(msg);
>     return (obj);
> }
> ```

Any suggestions on having a catch-all `pop_expected` that's a wrapper around a `static_assert()` telling developers to come here and add  another version to support the error type they want to use?

Currently devs would see a "cannot find function" error when trying to compile their code.
___
*[src/modules/packetio/workers/dpdk/worker_controller.cpp, line 493 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MM3Ax2k1TSeWymxkOWQ:-MM3Ax2k1TSeWymxkOWR:b27ayva) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/modules/packetio/workers/dpdk/worker_controller.cpp#L493)):*
> ```cpp
>     if (maybe_interface) { return (*maybe_interface); }
> 
>     return (tl::make_unexpected(ENODEV));
> ```

Went with `ENODEV` since it felt closest to what we're trying to convey. But am open to suggestions for a better error code.

DerangedMonkeyNinja (changes requested):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/387)* status: all files reviewed, 4 unresolved discussions (waiting on @dmorton-spirent)

___
*[src/framework/message/serialized_message.hpp, line 246 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MMGUZhf3umQ-ozeFeBS:-MMGUZhgDVEdj6tqz16A:b-a0d0hy) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/framework/message/serialized_message.hpp#L246)):*
> ```cpp
> 
>     auto& front = msg.parts.front();
>     auto obj = *(reinterpret_cast<tl::expected<T, std::string>*>(zmq_msg_data(&front)));
> ```

The zmq message shouldn't actually have a `std::string` in it, so I don't think this reinterpret cast here is correct.  The zmq message only has a c string in it.
___
*[src/framework/message/serialized_message.hpp, line 249 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MM3ATO-Bzj-IU7Ww7kh:-MMGUntW651we4uLSIz8:bkzlopg) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/framework/message/serialized_message.hpp#L249)):*

Any suggestions on having a catch-all `pop_expected` that's a wrapper around a `static_assert()` telling developers to come here and add  another version to support the error type they want to use?

Currently devs would see a "cannot find function" error when trying to compile their code.
</blockquote></details>

I believe you could add a generic template to accomplish that.  I'd have to look up the resolution rules, but I believe more specific matches override less specific ones.  So, it would just need to be the most generic template you could construct?
___
*[src/modules/packetio/internal_api.hpp, line 128 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MMGVYPeEnuv8AKHMTxH:-MMGVYPf-vRFbaLXPzFs:b-h6w9r9) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/modules/packetio/internal_api.hpp#L128)):*
> ```cpp
> struct reply_interface
> {
>     tl::expected<interface::generic_interface, int> interface;
> ```

There is already a `reply_error` struct for returning an error code.  Please use that instead of `tl::expected` here.  `reply_interface` should just return an interface.
___
*[src/modules/packetio/workers/dpdk/worker_controller.cpp, line 493 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MM3Ax2k1TSeWymxkOWQ:-MMGWnNwC2GUNz1OSM64:b6cp4kz) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/modules/packetio/workers/dpdk/worker_controller.cpp#L493)):*

Went with `ENODEV` since it felt closest to what we're trying to convey. But am open to suggestions for a better error code.
</blockquote></details>

I'd suggest `EINVAL` which is just a generic way to say 'bad argument'. That's also consistent with the other getter type functions here.

dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/387)* status: 3 of 8 files reviewed, 3 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/framework/message/serialized_message.hpp, line 246 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MMGUZhf3umQ-ozeFeBS:-MMHk1Vx875Bp1GonN8Y:b-2sk1ff) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/framework/message/serialized_message.hpp#L246)):*

The zmq message shouldn't actually have a `std::string` in it, so I don't think this reinterpret cast here is correct.  The zmq message only has a c string in it.
</blockquote></details>

Thanks. Moot now as I've reverted this file.
___
*[src/framework/message/serialized_message.hpp, line 249 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MM3ATO-Bzj-IU7Ww7kh:-MMHKGNCAoeqnkzFmgwd:b-loebib) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/framework/message/serialized_message.hpp#L249)):*

I believe you could add a generic template to accomplish that.  I'd have to look up the resolution rules, but I believe more specific matches override less specific ones.  So, it would just need to be the most generic template you could construct?
</blockquote></details>

I tried 
```
template <typenameT, t ypename E>
pop_expected(serialized_message &msg) { static_assert(); }
```
and the compiler errored saying there were multiple definitions of `pop_expected`. Which I guess makes sense since this version could still match and you'd get two functions with the same signature.

moot now since these functions are no longer required for this feature.
___
*[src/modules/packetio/internal_api.hpp, line 128 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MMGVYPeEnuv8AKHMTxH:-MMHk-KuB-GfoBYFzSIT:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/modules/packetio/internal_api.hpp#L128)):*

There is already a `reply_error` struct for returning an error code.  Please use that instead of `tl::expected` here.  `reply_interface` should just return an interface.
</blockquote></details>

Done.
___
*[src/modules/packetio/workers/dpdk/worker_controller.cpp, line 493 at r1](https://reviewable.io/reviews/spirent/openperf/387#-MM3Ax2k1TSeWymxkOWQ:-MMHM8bq-tMnI2KUxT99:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/52f7a4d21d50459cb90a83b9873d674309626c22/src/modules/packetio/workers/dpdk/worker_controller.cpp#L493)):*

I'd suggest `EINVAL` which is just a generic way to say 'bad argument'. That's also consistent with the other getter type functions here.
</blockquote></details>

Done.

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/387#-:-MMaSp2F3bzFMhhi-_7K:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @dmorton-spirent)

___
*[src/modules/packetio/internal_client.cpp, line 178 at r2](https://reviewable.io/reviews/spirent/openperf/387#-MMaRud88LmZ9Ql03NdG:-MMaRud88LmZ9Ql03NdH:b-2spiz8) ([raw file](https://github.com/spirent/openperf/blob/c0a47f9dadd11240be7bca6ee6e6e1fc24a60003/src/modules/packetio/internal_client.cpp#L178)):*
> ```cpp
>     if (!reply) { return (tl::make_unexpected(reply.error())); }
> 
>     if (auto maybe_interface = std::get_if<reply_interface>(&reply.value())) {
> ```

nit: prefer `auto*` over `auto` when declaring a known pointer. This is something I picked up since the start of this project, so there are lots of places where it isn't done.  Still, it's good practice.

PR #396, opened 2020-11-26
Network module

abelov-spirent:
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/396)


abelov-spirent:
**Network Server**
Root endpoint: `/network/servers`
model example:
```
{
  "id": "",
  "port": 3357,
  "protocol": "tcp",
  "stats": {
    "bytes_received": 0,
    "bytes_sent": 0,
    "closed": 0,
    "connections": 0,
    "errors": 0
  }
}
```

**Network generator:**
Root endpoint: `/network/generators`
Model example:
```
{
  "id": "",
  "config": {
    "connections": 1,
    "ops_per_connection": 1,
    "read_size": 256,
    "reads_per_sec": 100,
    "write_size": 512,
    "writes_per_sec": 100,
    "target": {
      "host": "127.0.0.1",
      "port": 3357,
      "protocol": "tcp"
    }
  }
}
```

**Network generator result:**
Root endpoint: `/network/generator-results`
Model example:
```
{
    "id": "",
    "generator_id": "qwe",
    "active": true,
    "timestamp_first": "2020-11-26T07:31:55.228669Z",
    "timestamp_last": "2020-11-26T07:32:39.675623Z",
    "connection": {
      "attempted": 0,
      "closed": 0,
      "errors": 0,
      "successful": 0
    },
    "read": {
      "bytes_actual": 0,
      "bytes_target": 0,
      "io_errors": 0,
      "latency_max": 0,
      "latency_min": 0,
      "latency_total": 0,
      "ops_actual": 0,
      "ops_target": 0
    },
    "write": {
      "bytes_actual": 0,
      "bytes_target": 0,
      "io_errors": 0,
      "latency_max": 0,
      "latency_min": 0,
      "latency_total": 0,
      "ops_actual": 0,
      "ops_target": 0
    },
    "dynamic_results": {}
  }
```

DerangedMonkeyNinja (commented):
> Reviewed 44 of 147 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 44 of 147 files reviewed, 21 unresolved discussions (waiting on @abelov-spirent)

___
*[api/schema/v1/modules/network.yaml, line 413 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTvQuN7cMFW6JM2DwA:-MNTvQuN7cMFW6JM2DwB:bjvoln1) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/api/schema/v1/modules/network.yaml#L413)):*
> ```YAML
>       port:
>         type: integer
>         format: int64
> ```

Suggest using `int32` instead of `int64` here as well as min/max values to limit the acceptable port range.
___
*[api/schema/v1/modules/network.yaml, line 433 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTvo433auelWXztb0x:-MNTvo433auelWXztb0y:breu586) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/api/schema/v1/modules/network.yaml#L433)):*
> ```YAML
>     type: object
>     description: Network server statistics
>     properties:
> ```

Please specify `minimum: 0` for all result counters.
___
*[api/schema/v1/modules/network.yaml, line 492 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTvw374ncWZnm-T9g_:-MNTvw374ncWZnm-T9ga:benpsw9) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/api/schema/v1/modules/network.yaml#L492)):*
> ```YAML
>           port:
>             type: integer
>             format: int64
> ```

Same comment as above.
___
*[src/modules/network/api_converters.hpp, line 9 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTwoTv6kmyuZbr_sAa:-MNTwoTv6kmyuZbr_sAb:bm2152e) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/api_converters.hpp#L9)):*
> ```cpp
> #include "api.hpp"
> 
> #include "swagger/v1/model/BulkCreateNetworkServersRequest.h"
> ```

I'd prefer for these headers to go in the cpp files instead of headers where they aren't explicitly needed.  I know it's easier to just include them all here, but it does make more work for the compiler and is bad practice.
___
*[src/modules/network/api_converters.cpp, line 34 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTxIdA-sy3ozR0xkRy:-MNTxIdA-sy3ozR0xkRz:b45d9hz) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/api_converters.cpp#L34)):*
> ```cpp
>         return "tcp";
>     default:
>         return "";
> ```

Suggest using an easier to identify error string such as 'unknown' or 'none'.
___
*[src/modules/network/api_converters.cpp, line 118 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTxhgL0f_mRIGTSKPz:-MNTxhgL0f_mRIGTSKQ-:brlsqjr) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/api_converters.cpp#L118)):*
> ```cpp
> {
>     request::generator::bulk::create request;
>     for (const auto& item : p_request.getItems())
> ```

While for loops are functional, I'd prefer the code operate at a higher level and use STL functions, e.g. `std::transform` here and `std::copy` below.
___
*[src/modules/network/generator_stack.cpp, line 58 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU-5YIBD1na9w2-UFs:-MNU-5YIBD1na9w2-UFt:bgw4c3g) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/generator_stack.cpp#L58)):*
> ```cpp
>         return tl::make_unexpected("Result not found");
>     }
>     return tl::make_unexpected("Result implemented");
> ```

I don't understand this response string.  Could it be made more specific?
___
*[src/modules/network/internal_utils.hpp, line 21 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU65c0DVx01TyATGrN:-MNU65c0DVx01TyATGrO:bqmb9gq) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/internal_utils.hpp#L21)):*
> ```cpp
> namespace openperf::network::internal {
> 
> union network_sockaddr
> ```

Please don't use unions in c++ code.  Use `std::variant` instead.
___
*[src/modules/network/internal_utils.hpp, line 61 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU8WAc9jOSH-CCTsM8:-MNU8WAc9jOSH-CCTsM9:bvrggeh) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/internal_utils.hpp#L61)):*
> ```cpp
> int ipv6_get_neighbor_ifindex(const in6_addr* addr);
> 
> /* These shouldn't be used directly.  Need to expose them for unit tests. */
> ```

If these are only needed for unit tests, then they belong with the unit test code.  Do *not* include anything that calls binaries from the main openperf code.
___
*[src/modules/network/internal_utils.cpp, line 126 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU6Sml107Rrapbnnsa:-MNU6Sml107Rrapbnnsb:b-d46w5i) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/internal_utils.cpp#L126)):*
> ```cpp
>  * Return the peer of the given socket in a freshly allocated string.
>  */
> char* get_peer_address(int socket)
> ```

Yikes!  I realize this is ported code, but...  Please update to a C++ equivalent.  There is no reason to manually allocate a string when you can just return `std::string`.

Also, a lot of these functions can be easily implemented using the net types in `src/lib/packet/types`.
___
*[src/modules/network/internal_utils_linux.cpp, line 63 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU71i57x5xOk9evZP-:-MNU71i57x5xOk9evZP0:b-10w8d3) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/internal_utils_linux.cpp#L63)):*
> ```cpp
>              *(ptr + 5));
> 
>     return mac_addr;
> ```

Yikes!  Please don't manually allocate strings in c++.
___
*[src/modules/network/drivers/dpdk.hpp, line 9 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU1v8M44pedqb_A0GU:-MNU1v8M44pedqb_A0GV:b-pmdvu5) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/drivers/dpdk.hpp#L9)):*
> ```cpp
> namespace openperf::network::internal::drivers {
> 
> class dpdk : public driver
> ```

This class should be marked as `final`.
___
*[src/modules/network/drivers/driver.hpp, line 18 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU0JuuA7JazmZDLkRf:-MNU0JuuA7JazmZDLkRg:b-lrrxr8) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/drivers/driver.hpp#L18)):*
> ```cpp
>     template <typename T> static std::shared_ptr<driver> instance()
>     {
>         static std::shared_ptr<driver> s_instance = nullptr;
> ```

Suggest inheriting from the `singleton` object in `src/framework/utils/singleton.hpp`.  That should be usable if you move the contents of `init` into the object constructor.

What's the rationale for using a shared pointer here?
___
*[src/modules/network/drivers/kernel.hpp, line 8 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU56UD-HrOzoMNdtbr:-MNU56UD-HrOzoMNdtbs:bq628ui) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/drivers/kernel.hpp#L8)):*
> ```cpp
> namespace openperf::network::internal::drivers {
> 
> class kernel : public driver
> ```

Please mark this class as `final`.
___
*[src/modules/network/firehose/protocol.hpp, line 20 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU2yvJ60Fpu34q-gxK:-MNU2yvJ60Fpu34q-gxL:b-5e8863) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/protocol.hpp#L20)):*
> ```cpp
> };
> 
> struct net_request_t
> ```

Suggest using `src/lib/packet/type/endian.hpp` based objects in your protocol struct.  They transparently handle endian issues.
___
*[src/modules/network/firehose/server_common.hpp, line 30 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU3JNT4rIxVFTMzXHM:-MNU3JNT4rIxVFTMzXHN:bt9rwnj) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/server_common.hpp#L30)):*
> ```cpp
>     size_t bytes_left;
>     std::vector<uint8_t> request;
>     union
> ```

Please use `std::variant` to replace c unions in c++ code.
___
*[src/modules/network/firehose/server_tcp.hpp, line 13 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU4y47CMLfGaDGfD3F:-MNU4y47CMLfGaDGfD3G:bq628ui) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/server_tcp.hpp#L13)):*
> ```cpp
> namespace openperf::network::internal::firehose {
> 
> class server_tcp : public server
> ```

Please mark this class as `final`.
___
*[src/modules/network/firehose/server_tcp.cpp, line 101 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU4Blf7g6Uvik3KDA3:-MNU4Blf7g6Uvik3KDA4:bb97211) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/server_tcp.cpp#L101)):*
> ```cpp
> server_tcp::server_tcp(in_port_t port, const drivers::driver_ptr& driver)
>     : m_stopped(false)
>     , m_context(zmq_ctx_new())
> ```

If you're going to spawn a new 0MQ context instead of using the existing system one, you need to disable the messaging threads.
___
*[src/modules/network/firehose/server_udp.hpp, line 13 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU4moQA7jMAOOMT8gk:-MNU4moQA7jMAOOMT8gl:bq628ui) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/server_udp.hpp#L13)):*
> ```cpp
> namespace openperf::network::internal::firehose {
> 
> class server_udp : public server
> ```

Please mark this class as `final`.
___
*[src/modules/network/models/server.hpp, line 35 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU9S1C4t5Lq_0XBivg:-MNU9S1C4t5Lq_0XBivh:b5u7fas) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/models/server.hpp#L35)):*
> ```cpp
>     virtual void id(std::string_view id) { m_id = id; }
> 
>     virtual int64_t port() const { return m_port; }
> ```

Network ports should use the `in_port_t` type.
___
*[src/modules/socket/directory.mk, line 14 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU7eu9-1BgSTcUA_zR:-MNU7eu9-1BgSTcUA_zS:b-c9zknj) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/socket/directory.mk#L14)):*
> ```Makefile
> SOCKSRV_SOURCES += \
> 	$(SOCK_COMMON) \
> 	client/api_client.cpp \
> ```

Please keep the client and server code separate.  I'd rather see a client library that can be optionally linked in instead of just munging all this code together.

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 20 of 149 files reviewed, 21 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/network.yaml, line 413 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTvQuN7cMFW6JM2DwA:-MNWgDr4-W8ZFrWpdyLb:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/api/schema/v1/modules/network.yaml#L413)):*

Suggest using `int32` instead of `int64` here as well as min/max values to limit the acceptable port range.
</blockquote></details>

Done.
___
*[api/schema/v1/modules/network.yaml, line 433 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTvo433auelWXztb0x:-MNWgO8m7sHE6zPuvj6u:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/api/schema/v1/modules/network.yaml#L433)):*

Please specify `minimum: 0` for all result counters.
</blockquote></details>

Done.
___
*[api/schema/v1/modules/network.yaml, line 492 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTvw374ncWZnm-T9g_:-MNWgWpT0Yl16O1vtRgS:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/api/schema/v1/modules/network.yaml#L492)):*

Same comment as above.
</blockquote></details>

Done.
___
*[src/modules/network/api_converters.hpp, line 9 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTwoTv6kmyuZbr_sAa:-MNWnddBE1Gm_SHskuqK:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/api_converters.hpp#L9)):*

I'd prefer for these headers to go in the cpp files instead of headers where they aren't explicitly needed.  I know it's easier to just include them all here, but it does make more work for the compiler and is bad practice.
</blockquote></details>

Done.
___
*[src/modules/network/api_converters.cpp, line 34 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTxIdA-sy3ozR0xkRy:-MNX5Sxv6LZeUuKQ9Bzp:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/api_converters.cpp#L34)):*

Suggest using an easier to identify error string such as 'unknown' or 'none'.
</blockquote></details>

Done.
___
*[src/modules/network/api_converters.cpp, line 118 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTxhgL0f_mRIGTSKPz:-MNWvpZd5tkTrijSjb95:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/api_converters.cpp#L118)):*

While for loops are functional, I'd prefer the code operate at a higher level and use STL functions, e.g. `std::transform` here and `std::copy` below.
</blockquote></details>

Done.
___
*[src/modules/network/generator_stack.cpp, line 58 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU-5YIBD1na9w2-UFs:-MNWwBPGDIMAwxIKCJEi:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/generator_stack.cpp#L58)):*

I don't understand this response string.  Could it be made more specific?
</blockquote></details>

Done.
___
*[src/modules/network/internal_utils.hpp, line 21 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU65c0DVx01TyATGrN:-MNXSlbJD-Vm-JyAA7PX:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/internal_utils.hpp#L21)):*

Please don't use unions in c++ code.  Use `std::variant` instead.
</blockquote></details>

Done.
___
*[src/modules/network/internal_utils.hpp, line 61 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU8WAc9jOSH-CCTsM8:-MNXXBpSBoXTfLjCZxiE:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/internal_utils.hpp#L61)):*

If these are only needed for unit tests, then they belong with the unit test code.  Do *not* include anything that calls binaries from the main openperf code.
</blockquote></details>

Done.
___
*[src/modules/network/drivers/dpdk.hpp, line 9 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU1v8M44pedqb_A0GU:-MNWhHqm-gTJrLYAMgNF:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/drivers/dpdk.hpp#L9)):*

This class should be marked as `final`.
</blockquote></details>

Done.
___
*[src/modules/network/drivers/driver.hpp, line 18 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU0JuuA7JazmZDLkRf:-MNZNoP4DY33wDPBnXI_:b-gcd8cg) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/drivers/driver.hpp#L18)):*

Suggest inheriting from the `singleton` object in `src/framework/utils/singleton.hpp`.  That should be usable if you move the contents of `init` into the object constructor.

What's the rationale for using a shared pointer here?
</blockquote></details>

Pointer is used here because driver is abstract. I was trying to apply utils::singleton here, but I did not find any 
___
*[src/modules/network/drivers/kernel.hpp, line 8 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU56UD-HrOzoMNdtbr:-MNWhJfz5J6nsfmX0r9Y:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/drivers/kernel.hpp#L8)):*

Please mark this class as `final`.
</blockquote></details>

Done.
___
*[src/modules/network/firehose/protocol.hpp, line 20 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU2yvJ60Fpu34q-gxK:-MNXtttu74wv-eE9OCxc:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/protocol.hpp#L20)):*

Suggest using `src/lib/packet/type/endian.hpp` based objects in your protocol struct.  They transparently handle endian issues.
</blockquote></details>

Done.
___
*[src/modules/network/firehose/server_tcp.hpp, line 13 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU4y47CMLfGaDGfD3F:-MNWhLOXEgKV15jbc1xp:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/server_tcp.hpp#L13)):*

Please mark this class as `final`.
</blockquote></details>

Done.
___
*[src/modules/network/firehose/server_tcp.cpp, line 101 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU4Blf7g6Uvik3KDA3:-MNZR-S95KRe6uBSHMhp:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/server_tcp.cpp#L101)):*

If you're going to spawn a new 0MQ context instead of using the existing system one, you need to disable the messaging threads.
</blockquote></details>

Done.
___
*[src/modules/network/firehose/server_udp.hpp, line 13 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU4moQA7jMAOOMT8gk:-MNWhMceEpgYRs7ZMSxm:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/server_udp.hpp#L13)):*

Please mark this class as `final`.
</blockquote></details>

Done.
___
*[src/modules/network/models/server.hpp, line 35 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU9S1C4t5Lq_0XBivg:-MNWgM_5AUdSMWT96pnT:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/models/server.hpp#L35)):*

Network ports should use the `in_port_t` type.
</blockquote></details>

Done.
___
*[src/modules/socket/directory.mk, line 14 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU7eu9-1BgSTcUA_zR:-MNZhio53NRp1Jk8xYj3:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/socket/directory.mk#L14)):*

Please keep the client and server code separate.  I'd rather see a client library that can be optionally linked in instead of just munging all this code together.
</blockquote></details>

Done.
___
*[src/modules/network/internal_utils.cpp, line 126 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU6Sml107Rrapbnnsa:-MNXXIZJEyZdSWtN5Hxy:b-wfytj3) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/internal_utils.cpp#L126)):*

Yikes!  I realize this is ported code, but...  Please update to a C++ equivalent.  There is no reason to manually allocate a string when you can just return `std::string`.

Also, a lot of these functions can be easily implemented using the net types in `src/lib/packet/types`.
</blockquote></details>

Removed as unused
Done.
___
*[src/modules/network/internal_utils_linux.cpp, line 63 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU71i57x5xOk9evZP-:-MNXXPidC-8YSsmir8o3:b-yvdoro) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/internal_utils_linux.cpp#L63)):*

Yikes!  Please don't manually allocate strings in c++.
</blockquote></details>

The same
Done.
___
*[src/modules/network/firehose/server_common.hpp, line 30 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNU3JNT4rIxVFTMzXHM:-MNY-S7OCvuK5L3GbBHr:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/src/modules/network/firehose/server_common.hpp#L30)):*

Please use `std::variant` to replace c unions in c++ code.
</blockquote></details>

Done.

DerangedMonkeyNinja (changes requested):
> Reviewed 14 of 147 files at r1, 29 of 37 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 62 of 150 files reviewed, 12 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/396#-MNz6TX5E181MHFCcs9K:-MNz6TX5E181MHFCcs9L:b90imef) (no related file):*
Why are the `src/modules/network/utils/ipv6_*.*` included at all?  They don't seem to have any code in them?

---
*[a discussion](https://reviewable.io/reviews/spirent/openperf/396#-MNz7vWt9jsRznzqADa2:-MNz7vWt9jsRznzqADa3:b-m0h1t3) (no related file):*
Can you add an AAT case to verify back-to-back operation with non-zero stats for both UDP and TCP over both IPv4 and IPv6?

---
___
*[api/schema/v1/modules/network.yaml, line 413 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTvQuN7cMFW6JM2DwA:-MNz1KMX2s1_5aeXQOj2:b-downfn) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/api/schema/v1/modules/network.yaml#L413)):*

Done.
</blockquote></details>

I don't see min/max here?
___
*[api/schema/v1/modules/network.yaml, line 492 at r1](https://reviewable.io/reviews/spirent/openperf/396#-MNTvw374ncWZnm-T9g_:-MNz1Sb4AeLic-qktJTl:b-ikcm8n) ([raw file](https://github.com/spirent/openperf/blob/417231b9e94b02925e7b44aea66cde16a4b699f2/api/schema/v1/modules/network.yaml#L492)):*

Done.
</blockquote></details>

Don't see min/max.
___
*[src/modules/network/task.cpp, line 84 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz36-84pCIMfZpA3Mq:-MNz36-84pCIMfZpA3Mr:b-6z8to3) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/task.cpp#L84)):*
> ```cpp
> 
> static int
> populate_sockaddr(network_sockaddr& s, const std::string& host, uint16_t port)
> ```

Nit: this should be `in_port_t` not `uint16_t`.
___
*[src/modules/network/task.cpp, line 87 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz2p2O2Dj2iifuF5zK:-MNz2p2O2Dj2iifuF5zL:b-o9op2p) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/task.cpp#L87)):*
> ```cpp
> {
>     sockaddr_storage client_storage;
>     auto sa4 = (sockaddr_in*)&client_storage;
> ```

Please use `auto*` when creating pointers. It helps your readers.

Additionally, try to limit variables to the scopes where they are actually use.  In this case, you don't need to define `sa4` or `sa6` except in the blocks where they are actually used.
___
*[src/modules/network/task.cpp, line 87 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz3G9Y1E8d2kVSOYCp:-MNz3G9Z-fNfefknAHMD:b-yjdygm) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/task.cpp#L87)):*
> ```cpp
> {
>     sockaddr_storage client_storage;
>     auto sa4 = (sockaddr_in*)&client_storage;
> ```

Please do not use `c` casts in `c++`.  Use `{static, reinterpret,const}_cast` as appropriate.
___
*[src/modules/network/task.cpp, line 188 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz3_GW67AUukFAf-KV:-MNz3_GW67AUukFAf-KW:b-9mygee) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/task.cpp#L188)):*
> ```cpp
>     int flags = m_driver->fcntl(sock, F_GETFL);
>     if (flags == -1) {
>         auto err = errno;
> ```

why not just `return tl::make_unexpected(errno)`?
___
*[src/modules/network/firehose/server_udp.cpp, line 123 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz4hCBB6dEfAhtekFz:-MNz4hCBB6dEfAhtekG-:b-zefsac) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/firehose/server_udp.cpp#L123)):*
> ```cpp
>         connection_t conn;
>         sockaddr_storage client_storage;
>         auto client = (sockaddr*)&client_storage;
> ```

Please don't use `c` casts in `c++` code.
___
*[src/modules/network/utils/network_sockaddr.hpp, line 48 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz5_Ty0hbEYju2UdH7:-MNz5_Ty0hbEYju2UdH8:b-watcg4) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/utils/network_sockaddr.hpp#L48)):*
> ```cpp
> }
> 
> inline void network_sockaddr_assign(const sockaddr* sa, network_sockaddr& ns)
> ```

Suggest returning a `network_sockaddr` instead of taking a reference to one as an argument.  It makes client code cleaner.
___
*[src/modules/network/utils/network_sockaddr.hpp, line 88 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz5leF4R3bGm3ST7mh:-MNz5leG4Hlt2XcMPkcG:b-zefsac) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/utils/network_sockaddr.hpp#L88)):*
> ```cpp
>     return std::visit(
>         utils::overloaded_visitor(
>             [&](const sockaddr_in& sa) { return (void*)&sa.sin_addr; },
> ```

Please don't use `c` casts in `c++` code.
___
*[src/modules/network/utils/ipv6.cpp, line 22 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz59E0BumYtx7dH9uU:-MNz59E0BumYtx7dH9uV:b-ony0ze) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/utils/ipv6.cpp#L22)):*
> ```cpp
> namespace openperf::network::internal {
> 
> bool ipv6_addr_is_link_local(const in6_addr* addr)
> ```

Why not use `libpacket::type::is_linklocal()`?

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 30 of 147 files reviewed, 11 unresolved discussions (waiting on @DerangedMonkeyNinja)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/396#-MNz6TX5E181MHFCcs9K:-MOHavpP1wtXtPEFaWpP:b-896fix) (no related file):*

Why are the `src/modules/network/utils/ipv6_*.*` included at all?  They don't seem to have any code in them?
</blockquote></details>

Done.

---
*[a discussion](https://reviewable.io/reviews/spirent/openperf/396#-MNz7vWt9jsRznzqADa2:-MOHawCK6w1Po5XVYRB0:b-896fix) (no related file):*

Can you add an AAT case to verify back-to-back operation with non-zero stats for both UDP and TCP over both IPv4 and IPv6?
</blockquote></details>

Done.

---
___
*[src/modules/network/task.cpp, line 87 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz2p2O2Dj2iifuF5zK:-MOAqbPc3mtkIxiPH64j:btpq4kk) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/task.cpp#L87)):*

Please use `auto*` when creating pointers. It helps your readers.

Additionally, try to limit variables to the scopes where they are actually use.  In this case, you don't need to define `sa4` or `sa6` except in the blocks where they are actually used.
</blockquote></details>

`sa4` and `sa6` are user at current block.
Done
___
*[src/modules/network/task.cpp, line 87 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz3G9Y1E8d2kVSOYCp:-MOAlvgk28lrGEHE97fY:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/task.cpp#L87)):*

Please do not use `c` casts in `c++`.  Use `{static, reinterpret,const}_cast` as appropriate.
</blockquote></details>

Done.
___
*[src/modules/network/task.cpp, line 188 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz3_GW67AUukFAf-KV:-MOAdlG71aBatElu9v33:bocygsn) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/task.cpp#L188)):*

why not just `return tl::make_unexpected(errno)`?
</blockquote></details>

`m_driver->close(sock);` change `errno` value
___
*[src/modules/network/firehose/server_udp.cpp, line 123 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz4hCBB6dEfAhtekFz:-MOAlnta6KmJ8iCVCzhy:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/firehose/server_udp.cpp#L123)):*

Please don't use `c` casts in `c++` code.
</blockquote></details>

Done.
___
*[src/modules/network/utils/network_sockaddr.hpp, line 48 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz5_Ty0hbEYju2UdH7:-MOBBb_90MUQHay0Epi8:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/utils/network_sockaddr.hpp#L48)):*

Suggest returning a `network_sockaddr` instead of taking a reference to one as an argument.  It makes client code cleaner.
</blockquote></details>

Done.
___
*[src/modules/network/utils/network_sockaddr.hpp, line 88 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz5leF4R3bGm3ST7mh:-MOAltYeEycNi9S63-JM:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/utils/network_sockaddr.hpp#L88)):*

Please don't use `c` casts in `c++` code.
</blockquote></details>

Done.
___
*[src/modules/network/utils/ipv6.cpp, line 22 at r2](https://reviewable.io/reviews/spirent/openperf/396#-MNz59E0BumYtx7dH9uU:-MOAu2qF-LYiNv2tDHyK:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/bcb1aa57ebe0ca340dfc70279d34f68bf59d9712/src/modules/network/utils/ipv6.cpp#L22)):*

Why not use `libpacket::type::is_linklocal()`?
</blockquote></details>

Done.

abelov-spirent:
dpdk driver is temporary disabled due to instability

DerangedMonkeyNinja (changes requested):
> Reviewed 32 of 43 files at r4.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 59 of 147 files reviewed, 9 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/network.yaml, line 426 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIV1gE7XfNHA_CiQ-v:-MOIV1gE7XfNHA_CiQ-w:bwypclm) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/api/schema/v1/modules/network.yaml#L426)):*
> ```YAML
>         type: string
>         description: Bind server socket to a particular device, specified as interface name
>       address_family:
> ```

This enum isn't necessary; please remove it. IP address formats are unambiguous.
___
*[api/schema/v1/modules/network.yaml, line 441 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIVIt5DQCNRZ9KoOQR:-MOIVIt5DQCNRZ9KoOQS:buvmtgl) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/api/schema/v1/modules/network.yaml#L441)):*
> ```YAML
>       - port
>       - protocol
>       - interface
> ```

I thought interface would be optional?  It's not required for IPv4 nor global IPv6, is it?
___
*[src/modules/network/firehose/server_tcp.cpp, line 36 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIYQkS3B_otjhTiAev:-MOIYQkS3B_otjhTiAew:b-59wk0t) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/firehose/server_tcp.cpp#L36)):*
> ```cpp
>     struct sockaddr_storage client_storage;
>     auto* server_ptr = reinterpret_cast<sockaddr*>(&client_storage);
>     std::string domain_str;
> ```

This variable makes my brain hurt.  Seems much simpler to just have a function return a string for the domain.  I'm pretty sure I saw one earlier.
___
*[src/modules/network/firehose/server_tcp.cpp, line 61 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIYh0DCEY3om4H8cck:-MOIYh0DCEY3om4H8ccl:b-fs3fp0) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/firehose/server_tcp.cpp#L61)):*
> ```cpp
>     int sock = 0, enable = true;
>     if ((sock = m_driver->socket(domain, SOCK_STREAM, IPPROTO_TCP)) == -1) {
>         OP_LOG(OP_LOG_WARNING,
> ```

This should be an error.
___
*[src/modules/network/firehose/server_tcp.cpp, line 137 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIYkys6lKOWdvAQi5Z:-MOIYkys6lKOWdvAQi5_:baujvs1) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/firehose/server_tcp.cpp#L137)):*
> ```cpp
>     if (domain) {
>         if ((res = new_server(domain.value(), port, interface)); res) {
>             OP_LOG(OP_LOG_INFO,
> ```

These INFO messages should be DEBUG messages.
___
*[src/modules/network/firehose/server_udp.cpp, line 106 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIZ9I5DWiu1dhBRtJj:-MOIZ9I5DWiu1dhBRtJk:b-xupbex) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/firehose/server_udp.cpp#L106)):*
> ```cpp
>     if (domain) {
>         if ((res = new_server(domain.value(), port, interface)); res) {
>             OP_LOG(OP_LOG_INFO,
> ```

Please make these debug messages.
___
*[src/modules/network/models/server.hpp, line 11 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIZNj94r3-PojH9D6i:-MOIZNj94r3-PojH9D6j:bm0cvjj) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/models/server.hpp#L11)):*
> ```cpp
> namespace openperf::network::model {
> 
> enum protocol_t : uint8_t { TCP, UDP };
> ```

Suggest using the standard values for protocol and family, e.g. IPPROTO_TCP, IPPROTO_UDP, AF_INET, AF_INET6.  Then at least you can use them directly in sockaddr structures without having to use switch statements.
___
*[src/modules/socket/directory.mk, line 31 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOI_ND_D9semK8ZHP6T:-MOI_ND_D9semK8ZHP6U:blc9eyg) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/socket/directory.mk#L31)):*
> ```Makefile
> 	server/udp_socket.cpp
> 
> ifeq ($(PLATFORM),linux)
> ```

This shouldn't be a part of your changelist...  Bad merge?
___
*[tests/aat/spec/network_spec.py, line 695 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOI_sT_8JEVYnAEdJdN:-MOI_sTa4kKpZy7LLxZh:b-fpfqc3) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/tests/aat/spec/network_spec.py#L695)):*
> ```Python
>                             result = self._api.get_network_generator_result(self._result.id)
>                             expect(result).to(be_valid_network_generator_result)
>                             try:
> ```

This doesn't seem like a useful test...  I must be reading it wrong, but doesn't that "failed with empty statistics" fire every time?

Also, could you please compare server and client stats.  The amount transmitted and received across a TCP link should be equal.  You can apply a fudge factor for the UDP version.  See `dataplane_spec.py` for an example of what I'm talking about.

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 45 of 147 files reviewed, 9 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/network.yaml, line 426 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIV1gE7XfNHA_CiQ-v:-MOUntAQ66EdmqxV8ROx:b-82iei4) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/api/schema/v1/modules/network.yaml#L426)):*

This enum isn't necessary; please remove it. IP address formats are unambiguous.
</blockquote></details>

We do not require ip address for server side, but dpdk for some reason does not work with IPv4 addresses on inet6 (also it does not return an error). So address_family is required for dpdk driver
___
*[api/schema/v1/modules/network.yaml, line 441 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIVIt5DQCNRZ9KoOQR:-MOUoRKF-ZvyRmpny80u:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/api/schema/v1/modules/network.yaml#L441)):*

I thought interface would be optional?  It's not required for IPv4 nor global IPv6, is it?
</blockquote></details>

Done.
___
*[src/modules/network/firehose/server_tcp.cpp, line 36 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIYQkS3B_otjhTiAev:-MOUpRi8-l0J5elthCgs:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/firehose/server_tcp.cpp#L36)):*

This variable makes my brain hurt.  Seems much simpler to just have a function return a string for the domain.  I'm pretty sure I saw one earlier.
</blockquote></details>

Done.
___
*[src/modules/network/firehose/server_tcp.cpp, line 61 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIYh0DCEY3om4H8cck:-MOUpag2EByIvKP_mFn5:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/firehose/server_tcp.cpp#L61)):*

This should be an error.
</blockquote></details>

Done.
___
*[src/modules/network/firehose/server_tcp.cpp, line 137 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIYkys6lKOWdvAQi5Z:-MOUpkQn162Nrq5BWZkU:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/firehose/server_tcp.cpp#L137)):*

These INFO messages should be DEBUG messages.
</blockquote></details>

Done.
___
*[src/modules/network/firehose/server_udp.cpp, line 106 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIZ9I5DWiu1dhBRtJj:-MOUpmPP4DiQUeCg5ZuT:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/firehose/server_udp.cpp#L106)):*

Please make these debug messages.
</blockquote></details>

Done.
___
*[src/modules/network/models/server.hpp, line 11 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIZNj94r3-PojH9D6i:-MOUrkDl4ianth6XjAEG:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/network/models/server.hpp#L11)):*

Suggest using the standard values for protocol and family, e.g. IPPROTO_TCP, IPPROTO_UDP, AF_INET, AF_INET6.  Then at least you can use them directly in sockaddr structures without having to use switch statements.
</blockquote></details>

Done.
___
*[src/modules/socket/directory.mk, line 31 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOI_ND_D9semK8ZHP6T:-MOUrzPc8CK4TnVfG4oF:bajptjl) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/socket/directory.mk#L31)):*

This shouldn't be a part of your changelist...  Bad merge?
</blockquote></details>

Socket module is required for network module. Both client and server. But if both  included at the same moment - `SOCKET COMMON` with be included 2 times. So my changes in `socket` module are required for `network` module to fix that conflict. 
___
*[tests/aat/spec/network_spec.py, line 695 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOI_sT_8JEVYnAEdJdN:-MOV0Ppz9lO0Mokfbka-:b-izxk5c) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/tests/aat/spec/network_spec.py#L695)):*

This doesn't seem like a useful test...  I must be reading it wrong, but doesn't that "failed with empty statistics" fire every time?

Also, could you please compare server and client stats.  The amount transmitted and received across a TCP link should be equal.  You can apply a fudge factor for the UDP version.  See `dataplane_spec.py` for an example of what I'm talking about.
</blockquote></details>

Fixed statistics check.
Current `framework/generator` does not support exact statistics values on stop signal. Posibly they will not include the last spin of the workers. I will discuss with Anatoly that issue, then after fix we can just enable that check for the network generator.

DerangedMonkeyNinja (changes requested):
> Reviewed 16 of 24 files at r5.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 60 of 148 files reviewed, 5 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/network.yaml, line 426 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIV1gE7XfNHA_CiQ-v:-MOfoAhp2Qosa8L9XTSM:bsrqatv) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/api/schema/v1/modules/network.yaml#L426)):*

We do not require ip address for server side, but dpdk for some reason does not work with IPv4 addresses on inet6 (also it does not return an error). So address_family is required for dpdk driver
</blockquote></details>

Then let's fix the bug in the DPDK/LwIP code and not pollute the API with a work around.
___
*[src/modules/network/firehose/server_tcp.cpp, line 177 at r5](https://reviewable.io/reviews/spirent/openperf/396#-MOfp6x2BSHRzsTavo3D:-MOfp6x2BSHRzsTavo3E:bn5l15d) ([raw file](https://github.com/spirent/openperf/blob/de4e4fbd15dce8ae6d180106badd78f10f129606/src/modules/network/firehose/server_tcp.cpp#L177)):*
> ```cpp
> void server_tcp::run_accept_thread()
> {
>     if (m_accept_thread.joinable()) {
> ```

This block is worrisome...  I assume you added it due to testing...

1) how is it triggered?
2) what can/should the user do if this is triggered?
___
*[src/modules/socket/client/api_client.hpp, line 46 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOfqdKE2B5QmOkQaWG3:-MOfqdKE2B5QmOkQaWG4:b-hs0j5x) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/socket/client/api_client.hpp#L46)):*
> ```cpp
>         io_channel_wrapper channel;
>     };
>     std::unordered_map<int, ided_channel> m_channels;
> ```

Why are you removing this from the client object?  These structures aren't thread safe, so if you use a static one, you can't have more than one thread/process EVER.  Please put it back.
___
*[tests/aat/spec/network_spec.py, line 695 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOI_sT_8JEVYnAEdJdN:-MOfrWG-8yQ0G9nCorEy:b-xm3tg3) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/tests/aat/spec/network_spec.py#L695)):*

Fixed statistics check.
Current `framework/generator` does not support exact statistics values on stop signal. Posibly they will not include the last spin of the workers. I will discuss with Anatoly that issue, then after fix we can just enable that check for the network generator.
</blockquote></details>

Isn't there a flag in the generator that explicitly tells you when the generator is finished?  Does that not work?

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 43 of 151 files reviewed, 4 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/network.yaml, line 426 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOIV1gE7XfNHA_CiQ-v:-MOklOXkBjtb5fAwDLpM:bd6msh9) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/api/schema/v1/modules/network.yaml#L426)):*

Then let's fix the bug in the DPDK/LwIP code and not pollute the API with a work around.
</blockquote></details>

Removed address_family from API. Issue with dpdk will be fixed in separate pull request
___
*[src/modules/network/firehose/server_tcp.cpp, line 177 at r5](https://reviewable.io/reviews/spirent/openperf/396#-MOfp6x2BSHRzsTavo3D:-MOj58Fz-3t-APPLak_D:bs626aa) ([raw file](https://github.com/spirent/openperf/blob/de4e4fbd15dce8ae6d180106badd78f10f129606/src/modules/network/firehose/server_tcp.cpp#L177)):*

This block is worrisome...  I assume you added it due to testing...

1) how is it triggered?
2) what can/should the user do if this is triggered?
</blockquote></details>

It will not pass aat tests if order will change.  In different order there is a chance zmq socket will throw segmentation fault
___
*[src/modules/socket/client/api_client.hpp, line 46 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOfqdKE2B5QmOkQaWG3:-MOklWXa40WSHsXMhHiY:b-b8w1aw) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/src/modules/socket/client/api_client.hpp#L46)):*

Why are you removing this from the client object?  These structures aren't thread safe, so if you use a static one, you can't have more than one thread/process EVER.  Please put it back.
</blockquote></details>

Changed to op_hashtab singleton
___
*[tests/aat/spec/network_spec.py, line 695 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOI_sT_8JEVYnAEdJdN:-MOj5YA2Dx7EGKaHFelW:b-lfh500) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/tests/aat/spec/network_spec.py#L695)):*

Isn't there a flag in the generator that explicitly tells you when the generator is finished?  Does that not work?
</blockquote></details>

Last statistics is just ignored by generator. Anatoly is working on fix right now (#429 )

DerangedMonkeyNinja (changes requested):
> Reviewed 23 of 31 files at r6.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 61 of 150 files reviewed, 6 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/generator.cpp, line 92 at r6](https://reviewable.io/reviews/spirent/openperf/396#-MOq7hsO5Jxr0LusqRfw:-MOq7hsP6UY0actdJb7k:b-9mi0z) ([raw file](https://github.com/spirent/openperf/blob/d595f11802f587137d8f2e31ac68936d3f38cfc6/src/modules/network/generator.cpp#L92)):*
> ```cpp
>             // +1 here is because target value has to be rounded to the upper
>             // value, otherwise it will be equal 0 on a first iteration
>             m_read_stat.ops_target = static_cast<uint64_t>(
> ```

Wouldn't the correct value be reads_per_sec * quanta when the elapsed time is 0?  Just adding 1 makes all future values off by 1, correct?
___
*[src/modules/network/firehose/server_tcp.cpp, line 177 at r5](https://reviewable.io/reviews/spirent/openperf/396#-MOfp6x2BSHRzsTavo3D:-MOq8K1610jIdpcKek_k:bv6tr0a) ([raw file](https://github.com/spirent/openperf/blob/de4e4fbd15dce8ae6d180106badd78f10f129606/src/modules/network/firehose/server_tcp.cpp#L177)):*

It will not pass aat tests if order will change.  In different order there is a chance zmq socket will throw segmentation fault
</blockquote></details>

That doesn't answer my question. How does this block affect ordering?
___
*[src/modules/socket/client/api_client.hpp, line 42 at r6](https://reviewable.io/reviews/spirent/openperf/396#-MOq8yWb2WfI7Q0LsvsB:-MOq8yWb2WfI7Q0LsvsC:b-wqgh6v) ([raw file](https://github.com/spirent/openperf/blob/d595f11802f587137d8f2e31ac68936d3f38cfc6/src/modules/socket/client/api_client.hpp#L42)):*
> ```cpp
>     unix_socket m_sock;
> 
>     channels_hashtab& m_channels;
> ```

Suggest just invoking the `instance()` method of the singleton when you need it.  No reason to store a local variable equal to a global one.
___
*[src/modules/socket/client/channels_hashtab.hpp, line 42 at r6](https://reviewable.io/reviews/spirent/openperf/396#-MOq6Osc3K_bRc0kJgUD:-MOq6OsdEUFNzcBYCQ0j:b2ptyn4) ([raw file](https://github.com/spirent/openperf/blob/d595f11802f587137d8f2e31ac68936d3f38cfc6/src/modules/socket/client/channels_hashtab.hpp#L42)):*
> ```cpp
>     }
> 
>     inline std::optional<ided_channel*> find(int fd)
> ```

Using `std::optional` for a pointer is redundant.  The pointer already has an implicitly unusable state, `nullptr`.
___
*[src/modules/socket/client/channels_hashtab.hpp, line 55 at r6](https://reviewable.io/reviews/spirent/openperf/396#-MOq6Gyr0c8koMQ18npm:-MOq6Gyr0c8koMQ18npn:b-ttb39g) ([raw file](https://github.com/spirent/openperf/blob/d595f11802f587137d8f2e31ac68936d3f38cfc6/src/modules/socket/client/channels_hashtab.hpp#L55)):*
> ```cpp
> 
> private:
>     op_hashtab* m_tab;
> ```

Suggest using a unique pointer for this.  Use C++ RAII facilities whenever possible.
___
*[tests/aat/spec/network_spec.py, line 695 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOI_sT_8JEVYnAEdJdN:-MOq9vXh7alH1e8lEXKR:bte0ho) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/tests/aat/spec/network_spec.py#L695)):*

Last statistics is just ignored by generator. Anatoly is working on fix right now (#429 )
</blockquote></details>

Add a comment here explaining this and what should happen after #429 is fixed.
Or, if #429 is merged before this is fixed, just go head and fix it. :)

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 55 of 151 files reviewed, 6 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/network/generator.cpp, line 92 at r6](https://reviewable.io/reviews/spirent/openperf/396#-MOq7hsO5Jxr0LusqRfw:-MP282k92iFbs5JOtlkH:bagbkts) ([raw file](https://github.com/spirent/openperf/blob/d595f11802f587137d8f2e31ac68936d3f38cfc6/src/modules/network/generator.cpp#L92)):*

Wouldn't the correct value be reads_per_sec * quanta when the elapsed time is 0?  Just adding 1 makes all future values off by 1, correct?
</blockquote></details>

First update will come after first worker iteration. So when the elapsed time is 0, statistics will contain correct values
___
*[src/modules/network/firehose/server_tcp.cpp, line 177 at r5](https://reviewable.io/reviews/spirent/openperf/396#-MOfp6x2BSHRzsTavo3D:-MP2KLBN94S8xrJ0RJu9:b-sjbfzd) ([raw file](https://github.com/spirent/openperf/blob/de4e4fbd15dce8ae6d180106badd78f10f129606/src/modules/network/firehose/server_tcp.cpp#L177)):*

That doesn't answer my question. How does this block affect ordering?
</blockquote></details>

Done.
Changed warning to assert with comment
Also fixed ordering of `accept` and `worker` threads
___
*[src/modules/socket/client/api_client.hpp, line 42 at r6](https://reviewable.io/reviews/spirent/openperf/396#-MOq8yWb2WfI7Q0LsvsB:-MP2AyNAAVbzXNWF5amF:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/d595f11802f587137d8f2e31ac68936d3f38cfc6/src/modules/socket/client/api_client.hpp#L42)):*

Suggest just invoking the `instance()` method of the singleton when you need it.  No reason to store a local variable equal to a global one.
</blockquote></details>

Done.
___
*[src/modules/socket/client/channels_hashtab.hpp, line 42 at r6](https://reviewable.io/reviews/spirent/openperf/396#-MOq6Osc3K_bRc0kJgUD:-MP2CYqf0sert0Q4PkQc:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/d595f11802f587137d8f2e31ac68936d3f38cfc6/src/modules/socket/client/channels_hashtab.hpp#L42)):*

Using `std::optional` for a pointer is redundant.  The pointer already has an implicitly unusable state, `nullptr`.
</blockquote></details>

Done.
___
*[src/modules/socket/client/channels_hashtab.hpp, line 55 at r6](https://reviewable.io/reviews/spirent/openperf/396#-MOq6Gyr0c8koMQ18npm:-MP2E-yOCsoYBpht0VlD:bvh0x81) ([raw file](https://github.com/spirent/openperf/blob/d595f11802f587137d8f2e31ac68936d3f38cfc6/src/modules/socket/client/channels_hashtab.hpp#L55)):*

Suggest using a unique pointer for this.  Use C++ RAII facilities whenever possible.
</blockquote></details>

`op_hashtab` is a C structure and it is used only as functions argument. `unique_ptr` may be redundant here
___
*[tests/aat/spec/network_spec.py, line 695 at r4](https://reviewable.io/reviews/spirent/openperf/396#-MOI_sT_8JEVYnAEdJdN:-MP2Hhlp6fYxSoIB1H18:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/a778dc84f3175c58cf94559627163205a3d5981f/tests/aat/spec/network_spec.py#L695)):*

Add a comment here explaining this and what should happen after #429 is fixed.
Or, if #429 is merged before this is fixed, just go head and fix it. :)
</blockquote></details>

Done.

src/modules/socket/server/socket_utils.cpp
```diff
@@ -88,10 +88,13 @@ std::optional<ip_addr_t> get_address(const sockaddr_storage& sstorage)
     }
     case AF_INET6: {
         auto sa6 = reinterpret_cast<const sockaddr_in6*>(&sstorage);
+        if (IN6_IS_ADDR_UNSPECIFIED(&sa6->sin6_addr)) {
```
cmyers-spirent, line 91: I guess this would change it so that IPV6_V6ONLY socket option is always treated as false.  Previously it seems it was always true.

I think it might be better to add a v6only option to this function and only do this when NOT v6only.

Calling functions would get the v6only flag from the PCB something like this:
addr = get_address(storage, IP_IS_V6_VAL(pcb->local_ip));

cmyers-spirent (commented):
> Reviewed 83 of 147 files at r1, 4 of 24 files at r5, 10 of 31 files at r6, 15 of 19 files at r7.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: 153 of 157 files reviewed, 7 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[tests/aat/spec/test_dataplane.yaml, line 23 at r7](https://reviewable.io/reviews/spirent/openperf/396#-MPLCT605UxFwj42PaFs:-MPLCT605UxFwj42PaFt:b-p9e7wf) ([raw file](https://github.com/spirent/openperf/blob/ac602e3a6a75435c0391d450eb307e10df840e75/tests/aat/spec/test_dataplane.yaml#L23)):*
> ```YAML
> 
>   network:
>     driver: dpdk
> ```

Is there a default when network driver is not set?
Maybe it should default to DPDK when DPDK is enabled?

abelov-spirent (commented):
> I guess this would change it so that IPV6_V6ONLY socket option is always treated as false.  Previously it seems it was always true.
> I think it might be better to add a v6only option to this function and only do this when NOT v6only.
> Calling functions would get the v6only flag from the PCB something like this:
> addr = get_address(storage, IP_IS_V6_VAL(pcb->local_ip));

Done

> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396#-:-MPO8DNZAq9sZLPkZg4B:blnsyqd)* status: 152 of 161 files reviewed, 7 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[tests/aat/spec/test_dataplane.yaml, line 23 at r7](https://reviewable.io/reviews/spirent/openperf/396#-MPLCT605UxFwj42PaFs:-MPNTfiN8hH6rqV-rU1R:b-nm9us6) ([raw file](https://github.com/spirent/openperf/blob/ac602e3a6a75435c0391d450eb307e10df840e75/tests/aat/spec/test_dataplane.yaml#L23)):*

Is there a default when network driver is not set?
Maybe it should default to DPDK when DPDK is enabled?
</blockquote></details>

I guess KERNEL driver as a default is more common and safe at this situation because it does not depend on other options. When you enabling dpdk driver at network, you will also check that dpdk is enabled as packetio module.
Switching default option values depending on some other options may be not a good practice.

cmyers-spirent (approved):
> Reviewed 7 of 147 files at r1, 5 of 37 files at r2, 1 of 43 files at r4, 4 of 31 files at r6, 2 of 19 files at r7, 9 of 9 files at r8.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/396)* status: all files reviewed, 6 unresolved discussions (waiting on @DerangedMonkeyNinja)

PR #411, opened 2020-12-07
TVLP Shared Buffer for Memory Generators in profile series

### LINKED ISSUE #399: Initialize memory buffer during creation of TVLP
Currently, the memory buffer is being initialized when the TVLP is started and may also occur during playback and incur undesirable wait times that affect playback.  The buffer should be initialized when the TVLP is created and released when it is deleted.

loskutnikov-spirent:
PR improves TVLP Memory Generator running.
Changes allow to allocate memory buffer once at TVLP creation time and using this buffer for all generators in the series of TVLP profile.

Closes #399.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/411)

### DIFF
--- a/src/modules/memory/api_transmogrify.cpp
+++ b/src/modules/memory/api_transmogrify.cpp
@@ -75,7 +75,10 @@ serialized_msg serialize(api_reply&& msg)
                      return openperf::message::push(serialized, item.id)
                             || openperf::message::push(serialized,
                                                        item.is_running)
-                            || openperf::message::push(serialized, item.config)
+                            || openperf::message::push(
+                                serialized,
+                                std::make_unique<decltype(item.config)>(
+                                    std::move(item.config)))
                             || openperf::message::push(
                                 serialized, item.init_percent_complete);
                  },
@@ -201,7 +204,8 @@ tl::expected<api_reply, int> deserialize_reply(serialized_msg&& msg)
         reply::generator::item item{};
         item.id = openperf::message::pop_string(msg);
         item.is_running = openperf::message::pop<bool>(msg);
-        item.config = openperf::message::pop<decltype(item.config)>(msg);
+        item.config = std::move(*std::unique_ptr<decltype(item.config)>(
+            openperf::message::pop<decltype(item.config)*>(msg)));
         item.init_percent_complete = openperf::message::pop<int32_t>(msg);
         return item;
     }
--- a/src/modules/memory/buffer.cpp
+++ b/src/modules/memory/buffer.cpp
@@ -1,0 +1,165 @@
+#include "buffer.hpp"
+
+#include "framework/core/op_log.h"
+
+#include <algorithm>
+#include <cassert>
+#include <cstdlib>
+#include <cstring>
+#include <new>
+
+namespace openperf::memory::internal {
+
+// Constructors & Destructor
+buffer::buffer(std::size_t alignment)
+    : m_alignment(alignment)
+{}
+
+buffer::buffer(std::size_t size, std::size_t alignment)
+    : buffer(alignment)
+{
+    allocate(size);
+}
+
+buffer::buffer(const buffer& other) { *this = other; }
+
+buffer::buffer(buffer&& other) noexcept
+    : m_alignment(other.m_alignment)
+    , m_ptr(other.m_ptr)
+    , m_size(other.m_size)
+{
+    other.m_ptr = nullptr;
+    other.m_size = 0;
+}
+
+buffer::~buffer() { release(); }
+
+buffer& buffer::operator=(const buffer& other)
+{
+    if (this == &other) return *this;
+
+    release();
+    m_alignment = other.m_alignment;
+
+    if (other.m_size > 0) {
+        allocate(other.m_size);
+
+        // NOTE: if statement needs to supress NonNullParamChecker warning
+        if (m_ptr != nullptr) std::memcpy(m_ptr, other.m_ptr, m_size);
+    }
+
+    return *this;
+}
+
+// Methods : public
+void buffer::allocate(std::size_t size)
+{
+    assert(size > 0);
+    assert(m_ptr == nullptr);
+
+    if (m_alignment) return allocate(size, m_alignment.value());
+
+    OP_LOG(OP_LOG_TRACE, "Allocating buffer of %zu bytes\n", size);
+
+    auto ptr = std::malloc(size);
+    if (ptr == nullptr) {
+        OP_LOG(
+            OP_LOG_ERROR, "Failed to allocate %zu bytes memory buffer\n", size);
+        throw std::bad_alloc();
+    }
+
+    m_ptr = ptr;
+    m_size = size;
+}
+
+void buffer::resize(std::size_t size)
+{
+    assert(size >= 0);
+    if (size == m_size) return;
+
+    if (size == 0) return release();
+    if (m_size == 0) return allocate(size);
+    if (m_alignment) return resize(size, m_alignment.value());
+
+    OP_LOG(OP_LOG_TRACE, "Resizing buffer (%zu => %zu)\n", m_size, size);
+
+    auto ptr = std::realloc(m_ptr, size);
+    if (ptr == nullptr) {
+        OP_LOG(OP_LOG_ERROR,
+               "Failed to resize (%zu => %zu) memory buffer\n",
+               m_size,
+               size);
+        throw std::bad_alloc();
+    }
+
+    m_ptr = ptr;
+    m_size = size;
+}
+
+void buffer::release()
+{
+    if (m_ptr == nullptr) return;
+
+    OP_LOG(OP_LOG_TRACE, "Releasing buffer of %zu bytes\n", m_size);
+
+    std::free(m_ptr);
+
+    m_ptr = nullptr;
+    m_size = 0;
+}
+
+// Methods : private
+void buffer::allocate(std::size_t size, std::size_t alignment)
+{
+    assert(size > 0);
+    assert(alignment > 0);
+    assert(m_ptr == nullptr);
+
+    OP_LOG(OP_LOG_TRACE,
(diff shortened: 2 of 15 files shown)


DerangedMonkeyNinja (changes requested):
> Reviewed 1 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/411)* status: 1 of 14 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja and @loskutnikov-spirent)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/411#-MNz-XMLEWRxgtcCcTlw:-MNz-XMLEWRxgtcCcTlx:bnob1oi) (no related file):*
A few comments:
1) Openperf isn't going to run under OSv, so there is no need to keep any concessions or comments related to OSv in the code base.
2) There is an existing aligned allocator for C++ in `src/framework/memory/aligned_allocator.hpp` which you can reuse.
3) All memory allocations are at least 8 byte aligned on 64 bit systems, whether you specify that or not.  I'd suggest making the alignment required so you can simplify the code.
4) If you need to keep both `malloc()` and `mmap()` based allocations, please just a template for your implementation.  That's not a runtime decision in your code, so don't write your implementation that way. 

---

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/411)* status: 1 of 15 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/411#-MNz-XMLEWRxgtcCcTlw:-MO4qiwZ64ZNoYVl7et-:b2btj6t) (no related file):*

A few comments:
1) Openperf isn't going to run under OSv, so there is no need to keep any concessions or comments related to OSv in the code base.
2) There is an existing aligned allocator for C++ in `src/framework/memory/aligned_allocator.hpp` which you can reuse.
3) All memory allocations are at least 8 byte aligned on 64 bit systems, whether you specify that or not.  I'd suggest making the alignment required so you can simplify the code.
4) If you need to keep both `malloc()` and `mmap()` based allocations, please just a template for your implementation.  That's not a runtime decision in your code, so don't write your implementation that way. 
</blockquote></details>

1. Removed.
2. I didn't see any advantages of using `aligned_allocator` against `std::aligned_alloc`. 
3. I believe explicit using of alignment is preferable.
4. `mmap()` functionality was removed.

---

DerangedMonkeyNinja (changes requested):
> Reviewed 7 of 14 files at r1, 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/411)* status: all files reviewed, 5 unresolved discussions (waiting on @loskutnikov-spirent)

___
*[src/modules/memory/buffer.cpp, line 62 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIMUdGDuyNj_ADZWJh:-MOIMUdH0ty9AIbdNZtm:b-88omfy) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/memory/buffer.cpp#L62)):*
> ```cpp
>     if (m_alignment) return allocate(size, m_alignment.value());
> 
>     OP_LOG(OP_LOG_INFO, "Allocating buffer of %zu bytes\n", size);
> ```

This (and all the INFO level messages below) should be at TRACE level.
___
*[src/modules/memory/generator.hpp, line 49 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIN2No8iEw_JVzvDJz:-MOIN2No8iEw_JVzvDK-:b-c6m76b) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/memory/generator.hpp#L49)):*
> ```cpp
> 
>     size_t m_buffer_size = 0;
>     std::shared_ptr<buffer> m_buffer;
> ```

Since size is a property of buffer now, do we really need `m_buffer_size` any more?
___
*[src/modules/memory/task_memory.cpp, line 76 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIObJEC1eoDlWx5C9R:-MOIObJEC1eoDlWx5C9S:bm9x8ca) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/memory/task_memory.cpp#L76)):*
> ```cpp
>     : m_pid(0.01, 0.0005, 0.0)
> {
>     if (auto size = sysconf(_SC_LEVEL1_DCACHE_LINESIZE); size > 0)
> ```

Please use a wrapper function to access this linux specific value.

Adding a common function in `core`, as part of a separate PR, would be super helpful.
___
*[src/modules/memory/task_memory_read.cpp, line 12 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIP7r5Au9ySeO8fiFh:-MOIP7r5Au9ySeO8fiFi:by6ivul) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/memory/task_memory_read.cpp#L12)):*
> ```cpp
>     assert(m_op_index < m_config.indexes->size());
> 
>     auto buffer_pointer = reinterpret_cast<uint8_t*>(m_buffer->data());
> ```

Nit: suggest using`auto* buffer_pointer` for this. It's a good signal to your readers than the variable is a pointer. Especially in non-hungarian naming cases...
___
*[src/modules/tvlp/workers/memory.cpp, line 19 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIPqLP2P27KkR-xn4J:-MOIPqLQ5VPL2RYerFUn:b-6sye5k) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/tvlp/workers/memory.cpp#L19)):*
> ```cpp
>     // Find the largest buffer from configuration
>     int64_t buffer_size = 0;
>     for (const auto& entry : series) {
> ```

nit: suggest `std::max_element`

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/411)* status: 12 of 15 files reviewed, 3 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/memory/buffer.cpp, line 62 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIMUdGDuyNj_ADZWJh:-MOKxCnQ7kSYRMwchOHD:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/memory/buffer.cpp#L62)):*

This (and all the INFO level messages below) should be at TRACE level.
</blockquote></details>

Done.
___
*[src/modules/memory/generator.hpp, line 49 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIN2No8iEw_JVzvDJz:-MOKxEaT0j0lsqwbTVmI:bk0zc0y) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/memory/generator.hpp#L49)):*

Since size is a property of buffer now, do we really need `m_buffer_size` any more?
</blockquote></details>

Yep, because an actual buffer size of a shared buffer passed to a generator may be greater than buffer size from generator configuration.
___
*[src/modules/memory/task_memory.cpp, line 76 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIObJEC1eoDlWx5C9R:-MOKxntsE4H0NXNIkdtB:b-tblepz) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/memory/task_memory.cpp#L76)):*

Please use a wrapper function to access this linux specific value.

Adding a common function in `core`, as part of a separate PR, would be super helpful.
</blockquote></details>

Ok, I will do it in separate PR, because CPU Generator uses such value too. 
___
*[src/modules/memory/task_memory_read.cpp, line 12 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIP7r5Au9ySeO8fiFh:-MOKy5-S0CkChSKij8ha:b-rbkhs9) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/memory/task_memory_read.cpp#L12)):*

Nit: suggest using`auto* buffer_pointer` for this. It's a good signal to your readers than the variable is a pointer. Especially in non-hungarian naming cases...
</blockquote></details>

Done. Hungarian names is nothing has with it, because the `_pointer` suffix was added to avoid confusing with the `buffer` type name.
___
*[src/modules/tvlp/workers/memory.cpp, line 19 at r2](https://reviewable.io/reviews/spirent/openperf/411#-MOIPqLP2P27KkR-xn4J:-MOKypVS1mbKbOVEHh_R:bbtfnmp) ([raw file](https://github.com/spirent/openperf/blob/ab42581c95f7d56fc9d8120c7dcb29a14f9ca3cb/src/modules/tvlp/workers/memory.cpp#L19)):*

nit: suggest `std::max_element`
</blockquote></details>

`std::max_element` returns the maximum element from an iterator. We need only maximum value of the field `buffer_size`. Additionally, custom comparator with internal type conversion from `json` to `swagger::MemoryGeneratorConfig` needs for this function, moreover returned value should be also converter to `swagger::MemoryGeneratorConfig` again. Seems too many excess code for extracting one field.

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/411#-:-MOfnufQ9TDyvB_6eVpm:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @loskutnikov-spirent)

PR #436, opened 2020-12-29
Network Modules performance fix

abelov-spirent:
- Fix UDP and TCP performance issue. O_NONBLOCK cause huge performance drop.
- Fix firehose protocol

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/436)

### DIFF
--- a/src/modules/network/drivers/dpdk.hpp
+++ b/src/modules/network/drivers/dpdk.hpp
@@ -17,6 +17,7 @@ public:
 
     void init() override;
     bool bind_to_device_required() override;
+    std::string_view driver_key() override { return key; };
 
     /* General socket functions */
     int accept(int s,
--- a/src/modules/network/drivers/dpdk.hpp
+++ b/src/modules/network/drivers/dpdk.hpp
@@ -17,7 +17,6 @@ public:
 
     void init() override;
     bool bind_to_device_required() override;
-    std::string_view driver_key() override { return key; };
 
     /* General socket functions */
     int accept(int s,
--- a/src/modules/network/drivers/driver.hpp
+++ b/src/modules/network/drivers/driver.hpp
@@ -26,6 +26,7 @@ public:
 
     virtual void init(){};
     virtual bool bind_to_device_required() { return false; };
+    virtual std::string_view driver_key() = 0;
 
     /* General socket functions */
     virtual int
--- a/src/modules/network/drivers/driver.hpp
+++ b/src/modules/network/drivers/driver.hpp
@@ -26,7 +26,7 @@ public:
 
     virtual void init(){};
     virtual bool bind_to_device_required() { return false; };
-    virtual std::string_view driver_key() = 0;
+    virtual bool support_socket_timeout() { return false; };
 
     /* General socket functions */
     virtual int
--- a/src/modules/network/drivers/kernel.hpp
+++ b/src/modules/network/drivers/kernel.hpp
@@ -13,7 +13,7 @@ public:
     kernel() = default;
     ~kernel() = default;
 
-    std::string_view driver_key() override { return key; };
+    bool support_socket_timeout() override { return true; };
 
     /* General socket functions */
     int accept(int s,
--- a/src/modules/network/drivers/kernel.hpp
+++ b/src/modules/network/drivers/kernel.hpp
@@ -13,6 +13,8 @@ public:
     kernel() = default;
     ~kernel() = default;
 
+    std::string_view driver_key() override { return key; };
+
     /* General socket functions */
     int accept(int s,
                struct sockaddr* addr,
--- a/src/modules/network/firehose/protocol.hpp
+++ b/src/modules/network/firehose/protocol.hpp
@@ -21,7 +21,7 @@ struct request_t
 struct net_request_t
 {
     char protocol[16];
-    libpacket::type::endian::field<1> action;
+    libpacket::type::endian::field<4> action;
     libpacket::type::endian::field<4> length;
 };
 const size_t net_request_size = sizeof(net_request_t);
@@ -37,7 +37,7 @@ inline tl::expected<request_t, int> parse_request(uint8_t*& data,
     }
 
     auto request = request_t{
-        .action = static_cast<action_t>(net_request->action.load<uint8_t>()),
+        .action = static_cast<action_t>(net_request->action.load<int32_t>()),
         .length = net_request->length.load<uint32_t>(),
     };
 
@@ -51,7 +51,7 @@ inline tl::expected<void, int> build_request(const request_t& request,
 
     strncpy(net_request.protocol, protocol, protocol_len);
 
-    net_request.action = request.action;
+    net_request.action.store<int32_t>(request.action);
     net_request.length = request.length;
 
     data.resize(sizeof(net_request));
--- a/src/modules/network/firehose/server_tcp.cpp
+++ b/src/modules/network/firehose/server_tcp.cpp
@@ -145,9 +145,7 @@ server_tcp::~server_tcp()
 {
     m_stopped.store(true, std::memory_order_relaxed);
 
-    // Closing blocked socket does not required for dpdk driver because of
-    // using SOCK_NONBLOCK flag
-    if (m_driver->driver_key().compare(drivers::dpdk::key) != 0) {
+    if (m_driver->support_socket_timeout()) {
         auto fd = m_blocked_socket.load(std::memory_order_relaxed);
         if (fd >= 0) {
             m_driver->shutdown(fd, SHUT_RDWR);
@@ -187,9 +185,7 @@ void server_tcp::run_accept_thread()
                     m_fd.load(),
                     client,
                     &client_length,
-                    (m_driver->driver_key().compare(drivers::dpdk::key) == 0)
-                        ? SOCK_NONBLOCK
-                        : 0))
+                    (!m_driver->support_socket_timeout()) ? SOCK_NONBLOCK : 0))
                != -1) {
             int enable = 1;
             if (m_driver->setsockopt(connect_fd,
@@ -202,10 +198,12 @@ void server_tcp::run_accept_thread()
                 continue;
             }
 
-            // SO_RCVTIMEO shows much better performance than SOCK_NONBLOCK
-            timeval tv = {.tv_sec = 0, .tv_usec = 1};
-            m_driver->setsockopt(
-                connect_fd, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv));
+            if (m_driver->support_socket_timeout()) {
+                // SO_RCVTIMEO shows much better performance than SOCK_NONBLOCK
+                timeval tv = {.tv_sec = 0, .tv_usec = 1};
+                m_driver->setsockopt(
+                    connect_fd, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv));
+            }
 
             auto conn = connection_msg_t{.fd = connect_fd};
             memcpy(&conn.client, client, get_sa_len(client));
--- a/src/modules/network/firehose/server_tcp.cpp
+++ b/src/modules/network/firehose/server_tcp.cpp
@@ -11,6 +11,7 @@
 #include "utils/random.hpp"
 #include "framework/utils/memcpy.hpp"
 
+#include "../drivers/dpdk.hpp"
 #include "protocol.hpp"
 
 namespace openperf::network::internal::firehose {
@@ -84,14 +85,6 @@ tl::expected<int, std::string> server_tcp::new_server(
         return tl::make_unexpected<std::string>(strerror(err));
     }
 
-    if (m_driver->setsockopt(
-            sock, SOL_SOCKET, SO_REUSEPORT, &enable, sizeof(enable))
-        != 0) {
-        auto err = errno;
-        m_driver->close(sock);
-        return tl::make_unexpected<std::string>(strerror(err));
-    }
-
     if (m_driver->bind(sock, server_ptr, get_sa_len(server_ptr)) == -1) {
         OP_LOG(OP_LOG_ERROR,
                "Unable to bind to socket (domain = %d, protocol = TCP): %s\n",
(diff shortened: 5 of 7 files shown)


cmyers-spirent (commented):
> Reviewed 5 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: 5 of 7 files reviewed, 3 unresolved discussions (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/firehose/server_tcp.cpp, line 86 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPglefT0qd4iwiPK-r9:-MPglefT0qd4iwiPK-rA:b-6ra15n) ([raw file](https://github.com/spirent/openperf/blob/7b7a6113ce6155377cfd02b7d604b1e041d8f9f8/src/modules/network/firehose/server_tcp.cpp#L86)):*
> ```cpp
>         return tl::make_unexpected<std::string>(strerror(err));
>     }
> ```

Why is the SO_REUSEPORT option being removed?
___
*[src/modules/network/firehose/server_tcp.cpp, line 147 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgwA7b2IDDvXjwoJVI:-MPgwA7b2IDDvXjwoJVJ:bh8mapm) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L147)):*
> ```cpp
> {
>     m_stopped.store(true, std::memory_order_relaxed);
> ```

This may not be thread-safe and could cause issues.  Is it really necessary?
Since you use SO_RECVTIMO, the thread should eventually exit when m_stopped =true.
___
*[src/modules/network/firehose/server_tcp.cpp, line 279 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgyOC77QAcjW4C8ydR:-MPgyOC77QAcjW4C8ydS:b-ty0qlu) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L279)):*
> ```cpp
>                 m_stat.connections++;
>             }
> ```

When processing multiple connections, using SO_RCVTIMEO on each socket it will impact performance if one connection has data and the other does not.

It would be better to use poll/epoll/select, but seems dpdk socket interface doesn't support that...

Is it common to have multiple connections?

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: 5 of 7 files reviewed, 3 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/firehose/server_tcp.cpp, line 86 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPglefT0qd4iwiPK-r9:-MPhm553BQuoJC7got2j:b-clafm9) ([raw file](https://github.com/spirent/openperf/blob/7b7a6113ce6155377cfd02b7d604b1e041d8f9f8/src/modules/network/firehose/server_tcp.cpp#L86)):*

Why is the SO_REUSEPORT option being removed?
</blockquote></details>

Accept thread is separate from workers so REUSEPORT is does not require in OpenPerf and may cause an issues
___
*[src/modules/network/firehose/server_tcp.cpp, line 147 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgwA7b2IDDvXjwoJVI:-MPhmUAL24Qid4fVaFq_:b-2lvacv) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L147)):*

This may not be thread-safe and could cause issues.  Is it really necessary?
Since you use SO_RECVTIMO, the thread should eventually exit when m_stopped =true.

</blockquote></details>

It is required to avoid block on write operations. Atomic here should guarantee of thread safety
___
*[src/modules/network/firehose/server_tcp.cpp, line 279 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgyOC77QAcjW4C8ydR:-MPhmwwDDSUqo4avCfCj:bhqu7ps) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L279)):*

When processing multiple connections, using SO_RCVTIMEO on each socket it will impact performance if one connection has data and the other does not.

It would be better to use poll/epoll/select, but seems dpdk socket interface doesn't support that...

Is it common to have multiple connections?
</blockquote></details>

Yes, it is common to have multiple connections. It gives much better performance than O_NONBLOCK.
DPDK does not support SO_RCVTIMEO anyway. And it has much better handling of O_NONBLOCK option. So it should be enouth.

cmyers-spirent (commented):
> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @abelov-spirent, @cmyers-spirent, and @DerangedMonkeyNinja)

___
*[src/modules/network/firehose/server_tcp.cpp, line 147 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgwA7b2IDDvXjwoJVI:-MPlcj_CEPgZOUcCZQyt:bml1uwx) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L147)):*

It is required to avoid block on write operations. Atomic here should guarantee of thread safety
</blockquote></details>

The acquiring the fd is atomic, but as far as I can tell DPDK sockets will not support close while writing to it.

The socket  API client socket channels are not reference counted and it is using a unprotected hashtab in the channels_hashtab.hpp and that might have some thread safety issues as well.  Seems like the openperf socket module may need some work...

Are there any known stability issues with DPDK based sockets?
___
*[src/modules/network/firehose/server_tcp.cpp, line 279 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgyOC77QAcjW4C8ydR:-MPljl3y96NuZuG47SVn:b1bk1do) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L279)):*

Yes, it is common to have multiple connections. It gives much better performance than O_NONBLOCK.
DPDK does not support SO_RCVTIMEO anyway. And it has much better handling of O_NONBLOCK option. So it should be enouth.
</blockquote></details>

Hmmm.  I guess as long as all connections are receiving packets are the same rate, this should get reasonable performance.
We probably will need to switch to poll/epoll/select eventually.

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/firehose/server_tcp.cpp, line 147 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgwA7b2IDDvXjwoJVI:-MPlpiqw0R_s7VsP37EN:bukzvhe) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L147)):*

The acquiring the fd is atomic, but as far as I can tell DPDK sockets will not support close while writing to it.

The socket  API client socket channels are not reference counted and it is using a unprotected hashtab in the channels_hashtab.hpp and that might have some thread safety issues as well.  Seems like the openperf socket module may need some work...

Are there any known stability issues with DPDK based sockets?
</blockquote></details>

O_NONBLOCK flag is used for dpdk driver, so close should not cause a stability issues. The easiest solution is to wait for worker thread stop and then to close sockets. Just for dpdk driver. What do you think?
___
*[src/modules/network/firehose/server_tcp.cpp, line 279 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgyOC77QAcjW4C8ydR:-MPoHV98ApCJndlFQjni:b-nnecz8) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L279)):*

Hmmm.  I guess as long as all connections are receiving packets are the same rate, this should get reasonable performance.
We probably will need to switch to poll/epoll/select eventually.

</blockquote></details>

I have tested performance of network server and current timeout bring much less delay than poll mode for kernel driver

cmyers-spirent (commented):
> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: all files reviewed, 1 unresolved discussion (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/firehose/server_tcp.cpp, line 147 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgwA7b2IDDvXjwoJVI:-MPoryVHAdSyOf4Z_Ood:bgofl61) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L147)):*

O_NONBLOCK flag is used for dpdk driver, so close should not cause a stability issues. The easiest solution is to wait for worker thread stop and then to close sockets. Just for dpdk driver. What do you think?
</blockquote></details>

Yeah I think it will be safer if dpdk driver mode is not able to perform socket operations on same fd from multiple threads.

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/firehose/server_tcp.cpp, line 147 at r1](https://reviewable.io/reviews/spirent/openperf/436#-MPgwA7b2IDDvXjwoJVI:-MPp7CZUErLJfVQzqpr5:bb6un1j) ([raw file](https://github.com/spirent/openperf/blob/2bbc4a65edf97035560631d5e58f9c515d3fe556/src/modules/network/firehose/server_tcp.cpp#L147)):*

Yeah I think it will be safer if dpdk driver mode is not able to perform socket operations on same fd from multiple threads.
</blockquote></details>

Done

cmyers-spirent (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: all files reviewed, 1 unresolved discussion (waiting on @abelov-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/firehose/server_tcp.cpp, line 149 at r2](https://reviewable.io/reviews/spirent/openperf/436#-MPpI5zSA6-GWnYdT3gX:-MPpI5zSA6-GWnYdT3gY:b-iwgi6o) ([raw file](https://github.com/spirent/openperf/blob/8f6d74e26f173938d3cafc6e8c69ee2d8e6800e8/src/modules/network/firehose/server_tcp.cpp#L149)):*
> ```cpp
> 
>     // Closing blocked socket does not required for dpdk driver because of
>     // using SOCK_NONBLOCK flag
> ```

since checking it in multiple places now would help to have a  function in the driver to check if blocking  socket read/write should be used e.g.:

  m_driver->needs_blocking_sockets()

abelov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: 2 of 6 files reviewed, all discussions resolved (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/firehose/server_tcp.cpp, line 149 at r2](https://reviewable.io/reviews/spirent/openperf/436#-MPpI5zSA6-GWnYdT3gX:-MPrUP0J1FZkbwHsjgGf:bb6un1j) ([raw file](https://github.com/spirent/openperf/blob/8f6d74e26f173938d3cafc6e8c69ee2d8e6800e8/src/modules/network/firehose/server_tcp.cpp#L149)):*

since checking it in multiple places now would help to have a  function in the driver to check if blocking  socket read/write should be used e.g.:

  m_driver->needs_blocking_sockets()
</blockquote></details>

Done

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @abelov-spirent and @cmyers-spirent)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/436#-MQCu0w61cbn1CtwEGIB:-MQCu0w61cbn1CtwEGIC:b-svmlmu) (no related file):*
Please revert the all of the conditional timeout code.  If the non-blocking socket is slower for the DPDK/socket implementation, then there is a bug in that code that needs to be fixed. So, let's fix that bug instead of working around it.

---

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/436)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/436#-MQCu0w61cbn1CtwEGIB:-MQEDJ4K0knhTruedSkv:bcik6md) (no related file):*

Please revert the all of the conditional timeout code.  If the non-blocking socket is slower for the DPDK/socket implementation, then there is a bug in that code that needs to be fixed. So, let's fix that bug instead of working around it.
</blockquote></details>

I think the work contract ended at the end of the year, so I mostly likely that means I will probably be working on this...
I haven't looked at the performance issue but according to Artem, the issue was only with the kernel driver.
Not sure why it would be slower, but constantly calling nonblocking recv() without waiting is not going to help anybody.
I suggested adding select/poll would be a better way to handle this, but there wasn't time to do that.

---

cmyers-spirent:
the network load generator was modified to use epoll (using event_loop) and the firehose protocol fix was added, so this pull request is not needed anymore.

PR #440, opened 2020-12-29
Improve AAT

loskutnikov-spirent:
Improves AAT tests for Block, Memory and CPU.
Unifies the Block, Memory and CPU test specifications.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/440)

### DIFF
--- a/src/framework/core/directory.mk
+++ b/src/framework/core/directory.mk
@@ -22,6 +22,7 @@ ifeq ($(PLATFORM), linux)
 	core/op_event_loop_epoll.c \
 	core/op_exit_backtrace.c \
 	core/op_thread_linux.c \
+	core/op_cpu_linux.c \
 	core/op_cpuset_linux.c
 else ifeq ($(PLATFORM), darwin)
 	FW_SOURCES += \
--- a/src/framework/core/op_cpu.h
+++ b/src/framework/core/op_cpu.h
@@ -1,0 +1,39 @@
+#ifndef _OP_CPU_HPP_
+#define _OP_CPU_HPP_
+
+#include <string.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ *  Get CPU architecture name
+ *
+ *  @return
+ *    The architecture name.
+ */
+const char* op_cpu_architecture(void);
+
+/**
+ *  Get CPU L1 cache line size
+ *
+ *  @return
+ *    The size in bytes or 0 if fail.
+ */
+uint16_t op_cpu_l1_cache_line_size(void);
+
+/**
+ * Get number of CPUs.
+ *
+ * @return
+ *   The number of CPUs or 0 if fail.
+ */
+size_t op_cpu_count(void);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // _OP_CPU_HPP_
--- a/src/framework/core/op_cpu_linux.c
+++ b/src/framework/core/op_cpu_linux.c
@@ -1,0 +1,23 @@
+#include "op_cpu.h"
+
+#include <unistd.h>
+#include <sys/utsname.h>
+
+uint16_t op_cpu_l1_cache_line_size()
+{
+    return sysconf(_SC_LEVEL1_DCACHE_LINESIZE);
+}
+
+const char* op_cpu_architecture()
+{
+    static struct utsname info;
+    uname(&info);
+    return info.machine;
+}
+
+size_t op_cpu_count(void)
+{
+    long count = sysconf(_SC_NPROCESSORS_ONLN);
+    if (count < 0) return 0;
+    return (size_t)count;
+}
--- a/src/framework/core/op_cpuset.c
+++ b/src/framework/core/op_cpuset.c
@@ -18,7 +18,7 @@ extern int op_cpuset_get_next(op_cpuset_t cpuset, size_t* cpu);
 op_cpuset_t op_cpuset_all(void)
 {
     op_cpuset_t cpuset = op_cpuset_create();
-    for (uint16_t core = 0; core < op_get_cpu_count(); core++) {
+    for (uint16_t core = 0; core < op_cpu_count(); core++) {
         op_cpuset_set(cpuset, core, true);
     }
 
--- a/src/framework/core/op_cpuset.h
+++ b/src/framework/core/op_cpuset.h
@@ -4,6 +4,8 @@
 #include <stddef.h>
 #include <stdbool.h>
 
+#include "op_cpu.h"
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -39,14 +41,6 @@ void op_cpuset_and(op_cpuset_t dest, op_cpuset_t src);
 void op_cpuset_or(op_cpuset_t dest, op_cpuset_t src);
 
 bool op_cpuset_equal(op_cpuset_t a, op_cpuset_t b);
-
-/**
- * Get number of cpus.
- *
- * @return
- *   The number of cpus or 0 if fail.
- */
-size_t op_get_cpu_count(void);
 
 #ifdef __cplusplus
 }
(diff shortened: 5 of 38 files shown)


cmyers-spirent (commented):
> Reviewed 38 of 38 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/440)* status: all files reviewed, 2 unresolved discussions (waiting on @DerangedMonkeyNinja and @loskutnikov-spirent)

___
*[src/framework/core/op_cpu.h, line 4 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPkJFLd3ozyirY75jHF:-MPkJFLd3ozyirY75jHG:bj2e3i4) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/src/framework/core/op_cpu.h#L4)):*
> ```objc
> #define _OP_CPU_HPP_
> 
> #include <string.h>
> ```

is string.h required?
___
*[src/modules/block/block_generator.cpp, line 203 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPl-LYK39pJ7A4PHKuG:-MPl-LYK39pJ7A4PHKuH:btr2gzf) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/src/modules/block/block_generator.cpp#L203)):*
> ```cpp
> 
> block_generator::block_result_ptr block_generator::statistics() const
> {
> ```

This looks a little more thread safe, but...

I think m_stat_ptr could be dereferenced when pointing to m_stat.
If memory copy is done while m_stat values are being updated there could be a data tear.

To be completely thread safe you would probably need to use RCU or maybe in this case a seqlock might be more appropriate.
Something like https://github.com/rigtorp/Seqlock

Seems like openperf mostly uses RCU.  I am not sure what Timmons would want here...
___
*[tests/aat/spec/memory_spec.py, line 500 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPkwxWL1lp9TmdvnaF7:-MPkwxWL1lp9TmdvnaF8:bbxy455) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/tests/aat/spec/memory_spec.py#L500)):*
> ```Python
>                     with before.all:
>                         for g7r in self._g8s:
>                             try:
> ```

Not really a good idea to have random values in unit tests for reproducibility reasons.
But I guess in this case ordering does not seem to be significant.

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/440)* status: all files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/framework/core/op_cpu.h, line 4 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPkJFLd3ozyirY75jHF:-MPmLBV03Rvxl1sTu8al:b-f6l3sv) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/src/framework/core/op_cpu.h#L4)):*

is string.h required?
</blockquote></details>

needs for `size_t` type alias
___
*[src/modules/block/block_generator.cpp, line 203 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPl-LYK39pJ7A4PHKuG:-MPmLSCFEEahZV8-DOuI:b-r5dxw7) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/src/modules/block/block_generator.cpp#L203)):*

This looks a little more thread safe, but...

I think m_stat_ptr could be dereferenced when pointing to m_stat.
If memory copy is done while m_stat values are being updated there could be a data tear.

To be completely thread safe you would probably need to use RCU or maybe in this case a seqlock might be more appropriate.
Something like https://github.com/rigtorp/Seqlock

Seems like openperf mostly uses RCU.  I am not sure what Timmons would want here...
</blockquote></details>

We used such approach in Memory and CPU early, was approved by Timmons. The OP has an intended lockless design and Timmons hasn't approves using locks.

We use RCU already here, because we copy the statistics, change the copy and than change pointer to new (updated) statistics data.
___
*[tests/aat/spec/memory_spec.py, line 500 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPkwxWL1lp9TmdvnaF7:-MPmO3ka45I1s7fBt8ao:bp4eogc) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/tests/aat/spec/memory_spec.py#L500)):*

Not really a good idea to have random values in unit tests for reproducibility reasons.
But I guess in this case ordering does not seem to be significant.
</blockquote></details>

The main goal of this acceptance test is checking the correctly start the set of generators in different states. We use `random` here to guarantee different states.

cmyers-spirent (approved):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/440)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @DerangedMonkeyNinja)

___
*[src/modules/block/block_generator.cpp, line 203 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPl-LYK39pJ7A4PHKuG:-MPmT6b3A4PtneoFy-lN:b-8gtpgn) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/src/modules/block/block_generator.cpp#L203)):*

We used such approach in Memory and CPU early, was approved by Timmons. The OP has an intended lockless design and Timmons hasn't approves using locks.

We use RCU already here, because we copy the statistics, change the copy and than change pointer to new (updated) statistics data.
</blockquote></details>

I don't think it is really RCU written like this because reader can have a reference to the pointer which is being modified.  For RCU you would need to use new memory which reader is guaranteed not to have a reference to.

Since the other generator stats are implemented the same way, I guess this change is OK.  I will check with Timmons on this when he gets back.
___
*[tests/aat/spec/memory_spec.py, line 500 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPkwxWL1lp9TmdvnaF7:-MPmUKGn7Dg0LrO8ZRCj:btjbua7) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/tests/aat/spec/memory_spec.py#L500)):*

The main goal of this acceptance test is checking the correctly start the set of generators in different states. We use `random` here to guarantee different states.
</blockquote></details>

for unit/AAT tests you normally want reproducibility and testing all paths through code, so random is usually a bad choice because if it fails due to a specific sequence it may not be reproducible again and each test run may not exercise all paths through code.

if using random sequences is guaranteed to test all paths through code every time then I think random is okay, but then again that would mean it doesn't need to be random either....

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/440)* status: 35 of 38 files reviewed, all discussions resolved (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[tests/aat/spec/memory_spec.py, line 500 at r1](https://reviewable.io/reviews/spirent/openperf/440#-MPkwxWL1lp9TmdvnaF7:-MPmoRpA0WvpMNB_tW0u:b-204afo) ([raw file](https://github.com/spirent/openperf/blob/beb415397f95d3faf1000f1afdfbe058ecacd5be/tests/aat/spec/memory_spec.py#L500)):*

for unit/AAT tests you normally want reproducibility and testing all paths through code, so random is usually a bad choice because if it fails due to a specific sequence it may not be reproducible again and each test run may not exercise all paths through code.

if using random sequences is guaranteed to test all paths through code every time then I think random is okay, but then again that would mean it doesn't need to be random either....
</blockquote></details>

The random was replaced by the odd/even item number of sequence.

PR #449, opened 2021-01-08
Refactor src gen 2

dmorton-spirent:
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/449)


dmorton-spirent (commented):
Thank you in advance for looking through a larger-than-normal PR!

This is mostly new changes with some minor updates to existing code.

> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449#-:-MQYiXhSBeLuDFXKK-Fb:b8wrgoc)* status: 0 of 42 files reviewed, 2 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/packet/generator/interface_source.cpp, line 240 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQYhsyV10lIB2_Jb7qh:-MQYhsyW6Boaz9ftaQ78:b-mbebo5) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L240)):*
> ```cpp
>     // clang.
>     // https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding
>     // std::for_each(sequence.begin(), sequence.end(), [&](auto seq) {
> ```

Will remove this line.
___
*[src/modules/packet/stack/lwip/include/lwipopts.h, line 109 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQYi3oZ3g4zKH5dqR2e:-MQYi3oZ3g4zKH5dqR2f:bavm9eu) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/stack/lwip/include/lwipopts.h#L109)):*
> ```objc
>  * Increase size of cache to a reasonable size.
>  */
> #define ARP_TABLE_SIZE 100
> ```

100 is completely arbitrary. Felt like a good compromise between having a reasonable cache size and not using too much additional memory.

DerangedMonkeyNinja (commented):
> Reviewed 1 of 42 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 1 of 42 files reviewed, 2 unresolved discussions (waiting on @DerangedMonkeyNinja and @dmorton-spirent)

___
*[src/modules/packet/stack/lwip/include/lwipopts.h, line 109 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQYi3oZ3g4zKH5dqR2e:-MQYpRVE8MAzUFjOJPND:b9p1wfz) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/stack/lwip/include/lwipopts.h#L109)):*

100 is completely arbitrary. Felt like a good compromise between having a reasonable cache size and not using too much additional memory.
</blockquote></details>

Suggest prefixing this variable with `LWIP_` for consistency.  Also suggest using a power of 2... They're just better! :)

dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 1 of 42 files reviewed, 2 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/packet/stack/lwip/include/lwipopts.h, line 109 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQYi3oZ3g4zKH5dqR2e:-MQZGDwjDjrbDZetnvX9:b-uewzzq) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/stack/lwip/include/lwipopts.h#L109)):*

Suggest prefixing this variable with `LWIP_` for consistency.  Also suggest using a power of 2... They're just better! :)
</blockquote></details>

I would add the `LWIP_` prefix, but am overriding an internal LwIP define here. Will bump it up to `256` though. Nice power of 2 and the magic size for a `/24`.

DerangedMonkeyNinja (changes requested):
> Reviewed 21 of 42 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 22 of 42 files reviewed, 12 unresolved discussions (waiting on @DerangedMonkeyNinja and @dmorton-spirent)

___
*[api/schema/v1/modules/packet/generator.yaml, line 503 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmgZGp3DNlT09ovOyv:-MQmgZGp3DNlT09ovOyw:b7op17b) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/api/schema/v1/modules/packet/generator.yaml#L503)):*
> ```YAML
>         description: |
>           Indicates whether this object is currently generating packets or not.
>       learning_resolved:
> ```

Would it be helpful to have more states here? Seems like the `learning_resolved = false` could be ambiguous since learning could have outright failed or just be "in progress". Additionally, port generators could just have `learning = none`, or something.
___
*[src/modules/packet/generator/interface_source.cpp, line 75 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmie-e5qfdb-FTuZC1:-MQmie-e5qfdb-FTuZC2:b-s4sl1b) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L75)):*
> ```cpp
>     // But lambda capture doesn't play well with structured binding under clang.
>     // https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding
>     for (auto seq : sequence) {
> ```

const?
___
*[src/modules/packet/generator/interface_source.cpp, line 190 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmiz3u7nDrY1vdxm2U:-MQmiz3u7nDrY1vdxm2V:b-7vlr70) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L190)):*
> ```cpp
> 
>     resolve_complete_callback callback =
>         std::bind(process_learning_results,
> ```

I think lambda's are generally preferred over `std::bind` nowadays.  At least, there is a clang-tidy option to enable warnings for `std::bind` usage.  Is there a reason you couldn't use a lambda here?
___
*[src/modules/packet/generator/interface_source.cpp, line 216 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmjMD422omL0m6gkQw:-MQmjMD5E0VIHwmj4szS:bzfrkld) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L216)):*
> ```cpp
> 
> // XXX: In an ideal world this would be constexpr.
> // But, the underlying std::array type doesn't zero out POD values.
> ```

I think this might depend on how you initialize the array.  I've definitely had std::array zeroing out data show up in profiling runs.
___
*[src/modules/packet/generator/interface_source.cpp, line 226 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmjeKo0kb2Q3qtnv1i:-MQmjeKo0kb2Q3qtnv1j:bfzn8y) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L226)):*
> ```cpp
>     libpacket::type::mac_address("00:00:00:00:00:00");
> 
> void interface_source::populate_source_addresses(traffic::sequence& sequence)
> ```

Given the similarities between this function and the one above, it might make sense to extract the get/set operation and turn them into a template.  Especially if the future IPv6 code ends up being nearly a duplicate as well.
___
*[src/modules/packet/generator/learning.cpp, line 108 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmk2Qr21Er4_wK0PY1:-MQmk2Qr21Er4_wK0PY2:b-ghn1x7) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/learning.cpp#L108)):*
> ```cpp
> {
>     // If we're already learning, don't start again.
>     if (in_progress()) { return (false); }
> ```

FYI: This is why I like to include a state tag struct with each event function.  This prevents all of these repetitive state checks.  

For example, does something call `start_learning_impl()` directly? If so, why? If not, why do we check for the `in_progress()` state in both cases?
___
*[src/modules/packet/generator/port_source.hpp, line 8 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmkfGTEwrv3IyhCeuP:-MQmkfGTEwrv3IyhCeuQ:bd3m7yk) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/port_source.hpp#L8)):*
> ```cpp
> {
> public:
>     static constexpr bool supports_learning = false;
> ```

Could this property be made the result of a function call?  This is just a mapping between a type and a value, correct?
___
*[src/modules/packet/generator/source.hpp, line 120 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmlIvY9YFrEx0Gnv_s:-MQmlIvY9YFrEx0Gnv_t:b-avumjt) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/source.hpp#L120)):*
> ```cpp
>     bool supports_learning() const;
> 
>     std::optional<bool> maybe_retry_learning();
> ```

optional<bool> seems silly.  You're turning a two state return value into a three state return value. Do you need all three states?
___
*[src/modules/packet/generator/source.cpp, line 96 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmla9z8KECLD2Ri4Cc:-MQmla9z8KECLD2Ri4Cd:btht9gr) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/source.cpp#L96)):*
> ```cpp
>     if (maybe_port_index.has_value()) { return (port_source{}); }
> 
>     auto maybe_interface = client.interface(target_id);
> ```

nit: you can simplify this code pattern here (and elsewhere)...
```
if (auto interface = client.interface(target_id)) {
    return (interface_source{loop, interface.value()});
}
```

`std::optional` is implicitly convertible to bool (like a pointer), so you can do the assignment and branch in just one line.
___
*[src/modules/packet/generator/traffic/sequence.hpp, line 32 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmmOto8aBXX9bYP_8P:-MQmmOto8aBXX9bYP_8Q:b-lfm3x9) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/traffic/sequence.hpp#L32)):*
> ```cpp
>     using const_iterator = const iterator;
> 
>     static constexpr size_t flow_index = static_cast<size_t>(0);
> ```

The casts here are redundant, aren't they?
___
*[tests/aat/spec/packet_generator_spec.py, line 370 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmpzHO4rkBFCFWgl-4:-MQmpzHO4rkBFCFWgl-5:br087lv) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/tests/aat/spec/packet_generator_spec.py#L370)):*
> ```Python
>                             expect(result).to(be_valid_packet_generator)
> 
>                             # XXX: learning state implicitly verified during start generator test.
> ```

Could you verify that the learned MAC addresses match the opposite interface?

DerangedMonkeyNinja (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 22 of 42 files reviewed, 12 unresolved discussions (waiting on @dmorton-spirent)

___
*[src/modules/packet/stack/lwip/include/lwipopts.h, line 109 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQYi3oZ3g4zKH5dqR2e:-MQmqQjX1tKXThxPU11x:b-b0j6en) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/stack/lwip/include/lwipopts.h#L109)):*

I would add the `LWIP_` prefix, but am overriding an internal LwIP define here. Will bump it up to `256` though. Nice power of 2 and the magic size for a `/24`.
</blockquote></details>

D'oh.  I see... nevermind.

dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 22 of 42 files reviewed, 11 unresolved discussions (waiting on @DerangedMonkeyNinja and @dmorton-spirent)

___
*[api/schema/v1/modules/packet/generator.yaml, line 503 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmgZGp3DNlT09ovOyv:-MQmwZf5AR_fKQM_PFLx:b-uidjc0) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/api/schema/v1/modules/packet/generator.yaml#L503)):*

Would it be helpful to have more states here? Seems like the `learning_resolved = false` could be ambiguous since learning could have outright failed or just be "in progress". Additionally, port generators could just have `learning = none`, or something.
</blockquote></details>

Good points. I did go back and forth on this.

My thought process:
* Boolean is something easy for clients to process.
* At this level it's all they really need to know. They can dive into the `/learning` endpoint if they want to know more.
* This is only included for interface generators. Port generators omit this property.
___
*[src/modules/packet/generator/interface_source.cpp, line 190 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmiz3u7nDrY1vdxm2U:-MQn0ZsODyHvtyBDOIOz:b-iwazkx) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L190)):*

I think lambda's are generally preferred over `std::bind` nowadays.  At least, there is a clang-tidy option to enable warnings for `std::bind` usage.  Is there a reason you couldn't use a lambda here?
</blockquote></details>

I didn't think to try :)? I'll definitely give that a shot, thx for the suggestion!
___
*[src/modules/packet/generator/interface_source.cpp, line 216 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmjMD422omL0m6gkQw:-MQmsMHZ2P9GTzgpzH9p:b-b2zntx) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L216)):*

I think this might depend on how you initialize the array.  I've definitely had std::array zeroing out data show up in profiling runs.
</blockquote></details>

Yeah, I noticed that in other places.

Citing my source: `note that default initialization may result in indeterminate values for non-class T` from https://en.cppreference.com/w/cpp/container/array. Since the code depends on this value always being all zeros, went with the approach that guarantees this. Else you could end up with random, intermittent cases where source addresses don't update.
___
*[src/modules/packet/generator/interface_source.cpp, line 226 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmjeKo0kb2Q3qtnv1i:-MQnNdOrEp3CMcEkaMV6:b-ryj9or) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L226)):*

Given the similarities between this function and the one above, it might make sense to extract the get/set operation and turn them into a template.  Especially if the future IPv6 code ends up being nearly a duplicate as well.
</blockquote></details>

With IPv6 support right around the corner, I'd like to defer this suggestion and implement it as part of that change. I will take a good, long look at it.
___
*[src/modules/packet/generator/learning.cpp, line 108 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmk2Qr21Er4_wK0PY1:-MQnOIIe535_3iyGuC6B:b-p1bteq) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/learning.cpp#L108)):*

FYI: This is why I like to include a state tag struct with each event function.  This prevents all of these repetitive state checks.  

For example, does something call `start_learning_impl()` directly? If so, why? If not, why do we check for the `in_progress()` state in both cases?
</blockquote></details>

Thanks for the tip!
___
*[src/modules/packet/generator/port_source.hpp, line 8 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmkfGTEwrv3IyhCeuP:-MQmtMbE3_HzjsgMTNRT:b-56yz64) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/port_source.hpp#L8)):*

Could this property be made the result of a function call?  This is just a mapping between a type and a value, correct?
</blockquote></details>

Yeah, this could be something like a free function that wraps an `overloaded_visitor` in `source.cpp`. My thought process had been to let the various source detail types self-identify as supporting learning or not. 6 of one, 1/2 dozen of the other. I'll switch to a function call in `source.cpp`. 
___
*[src/modules/packet/generator/source.hpp, line 120 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmlIvY9YFrEx0Gnv_s:-MQmukC05S3zk1sV65xj:b-eouuvo) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/source.hpp#L120)):*

optional<bool> seems silly.  You're turning a two state return value into a three state return value. Do you need all three states?
</blockquote></details>

Yes. At this point the caller has no idea if the underlying source type supports learning. So:
1. Optional value == true, start of learning retry successful.
2. Optional value == false, start of learning retry failed. (this signals an error)
3. Optional has no value, learning did not start because underlying source type doesn't support it. (not an error. or at the very least a different error from item 2.)
___
*[src/modules/packet/generator/traffic/sequence.hpp, line 32 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmmOto8aBXX9bYP_8P:-MQmvvLi8ub5KA7ajwnQ:b-c1kmcn) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/traffic/sequence.hpp#L32)):*

The casts here are redundant, aren't they?
</blockquote></details>

At one point `clang-tidy` was flagging this for not having them. It's since stopped and I've removed them.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 22 of 42 files reviewed, 5 unresolved discussions (waiting on @dmorton-spirent)

___
*[api/schema/v1/modules/packet/generator.yaml, line 503 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmgZGp3DNlT09ovOyv:-MQs7yZ6E8poaanfeTAr:bafdcnr) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/api/schema/v1/modules/packet/generator.yaml#L503)):*

Good points. I did go back and forth on this.

My thought process:
* Boolean is something easy for clients to process.
* At this level it's all they really need to know. They can dive into the `/learning` endpoint if they want to know more.
* This is only included for interface generators. Port generators omit this property.
</blockquote></details>

My counterpoints... :)

* It's just as easy for clients to check 'learning == completed'
* If we're still in the process of learning, then there's no need for clients to retrieve the `/learning` endpoint. They just need to wait.
___
*[src/modules/packet/generator/interface_source.cpp, line 216 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmjMD422omL0m6gkQw:-MQs8adz69RrmKBv44hI:b-gxztsp) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L216)):*

Yeah, I noticed that in other places.

Citing my source: `note that default initialization may result in indeterminate values for non-class T` from https://en.cppreference.com/w/cpp/container/array. Since the code depends on this value always being all zeros, went with the approach that guarantees this. Else you could end up with random, intermittent cases where source addresses don't update.
</blockquote></details>

Fair enough.
___
*[src/modules/packet/generator/source.hpp, line 120 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmlIvY9YFrEx0Gnv_s:-MQs91491JpywyaBvSOA:b-afjeqo) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/source.hpp#L120)):*

Yes. At this point the caller has no idea if the underlying source type supports learning. So:
1. Optional value == true, start of learning retry successful.
2. Optional value == false, start of learning retry failed. (this signals an error)
3. Optional has no value, learning did not start because underlying source type doesn't support it. (not an error. or at the very least a different error from item 2.)
</blockquote></details>

Ok...  But what behavioral difference in callers is there between 1 and 3. The caller has nothing left to do in either case, right?

Also, I forgot to mention earlier, returning an `enum` here would be clearer for readers. Try to limit boolean return values to explicit questions, e.g. `has_feature()` or `needs_thing_done()`.

djoyner (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 22 of 42 files reviewed, 5 unresolved discussions (waiting on @DerangedMonkeyNinja and @dmorton-spirent)

___
*[api/schema/v1/modules/packet/generator.yaml, line 503 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmgZGp3DNlT09ovOyv:-MQsDuK9DGsAdNPhEMhw:b-bdwu32) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/api/schema/v1/modules/packet/generator.yaml#L503)):*

My counterpoints... :)

* It's just as easy for clients to check 'learning == completed'
* If we're still in the process of learning, then there's no need for clients to retrieve the `/learning` endpoint. They just need to wait.
</blockquote></details>

I also tend to use state enums for this sort of thing. Booleans by themselves carry very little info and can be easily misinterpreted. Not as likely in this case given the field name, but still a good thing to consider (read https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/ if you're interested in learning more).

dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 22 of 42 files reviewed, 5 unresolved discussions (waiting on @DerangedMonkeyNinja and @dmorton-spirent)

___
*[api/schema/v1/modules/packet/generator.yaml, line 503 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmgZGp3DNlT09ovOyv:-MQsDOnA8mhwqfdw7ydn:bpud2ed) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/api/schema/v1/modules/packet/generator.yaml#L503)):*
<details><summary><i>Previously, djoyner (David Joyner) wrote…</i></summary><blockquote>

I also tend to use state enums for this sort of thing. Booleans by themselves carry very little info and can be easily misinterpreted. Not as likely in this case given the field name, but still a good thing to consider (read https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/ if you're interested in learning more).
</blockquote></details>

Great additional context, thanks!

I will switch to an enum string instead of a boolean here.
___
*[src/modules/packet/generator/source.hpp, line 120 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmlIvY9YFrEx0Gnv_s:-MQsHMEt50sq09TS_2Zc:b-e3jwyb) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/source.hpp#L120)):*

Ok...  But what behavioral difference in callers is there between 1 and 3. The caller has nothing left to do in either case, right?

Also, I forgot to mention earlier, returning an `enum` here would be clearer for readers. Try to limit boolean return values to explicit questions, e.g. `has_feature()` or `needs_thing_done()`.
</blockquote></details>

True. As of today callers treat items 1 and 3 the same. Differentiating them gives the caller the option of returning a warning about learning not supported vs a success message. I didn't want to "lie" to the caller and tell them learning started when in fact nothing happened.

Using an `enum` seems like the better solution here. Will switch to that.

dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 7 of 44 files reviewed, 4 unresolved discussions (waiting on @DerangedMonkeyNinja and @dmorton-spirent)

___
*[src/modules/packet/generator/interface_source.cpp, line 75 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmie-e5qfdb-FTuZC1:-MQyCEucDTymq-8fe5W8:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L75)):*

const?
</blockquote></details>

Done.
___
*[tests/aat/spec/packet_generator_spec.py, line 370 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmpzHO4rkBFCFWgl-4:-MQyCRBwEbVvfVHouXT6:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/tests/aat/spec/packet_generator_spec.py#L370)):*

Could you verify that the learned MAC addresses match the opposite interface?
</blockquote></details>

Done.

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 15 of 25 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449#-:-MR-yFCs094veFZmvl24:bnfp4nl)* status: 22 of 44 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja and @dmorton-spirent)

___
*[src/modules/packet/generator/interface_source.cpp, line 75 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmie-e5qfdb-FTuZC1:-MR-xGWP2VOsE2P_Sa9Q:b-1lem3u) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L75)):*

Done.
</blockquote></details>

reference? :)

dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/449)* status: 22 of 44 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/packet/generator/interface_source.cpp, line 75 at r1](https://reviewable.io/reviews/spirent/openperf/449#-MQmie-e5qfdb-FTuZC1:-MQnNatN12AKXrXIJZes:b-ksqel2) ([raw file](https://github.com/spirent/openperf/blob/421709c4435866967a565e02eeaa78802d722f68/src/modules/packet/generator/interface_source.cpp#L75)):*

reference? :)
</blockquote></details>

`tcp-lint` to the rescue! ;)

PR #457, opened 2021-01-15
modules/packetio: change pool allocation scheme

DerangedMonkeyNinja:
Update the memory pool allocation scheme to support two distinct types
of memory pools: shared and unique. Unique pools are used when we can be
confident that the source is going to overwrite the payload for each
transmitted packet. Unique pools are used when that doesn't happen.

This prevents the case where payload data from previously used packet
buffers appears in packets without any user generated payload.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/457)

### DIFF
--- a/src/modules/packetio/memory/dpdk/mempool.hpp
+++ b/src/modules/packetio/memory/dpdk/mempool.hpp
@@ -8,6 +8,8 @@ struct rte_mempool;
 
 namespace openperf::packetio::dpdk::mempool {
 
+enum class mempool_type { none = 0, unique, shared };
+
 /*
  * The next two functions are intended for acquiring/releasing memory pools for
  * transmit purposes.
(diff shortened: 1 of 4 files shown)


dmorton-spirent (approved):
:lgtm: 

> Reviewed 3 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/457#-:-MRWMjuG0HxVEJGeqMxg:bg7igkv)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/packetio/memory/dpdk/primary/mempool.cpp, line 288 at r1](https://reviewable.io/reviews/spirent/openperf/457#-MRWLP8T0Mekz2aOUIx7:-MRWLP8UA76wgC4dTPag:b6gi3un) ([raw file](https://github.com/spirent/openperf/blob/ae86821d08dfefa935b5fa55bf79ca5b9c2a050c/src/modules/packetio/memory/dpdk/primary/mempool.cpp#L288)):*
> ```cpp
>             create_mempool(numa_node, packet_length, packet_count, cache_size));
>     default:
>         /* mempool_type::none?  Ok... */
> ```

Nit: would it be helpful to add a debug print here to tell developers/users they're going to get a nullptr because they gave an unsupported `mempool_type`?

PR #459, opened 2021-01-23
Fix socket client channel/eventfd leak

cmyers-spirent:
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/459)

### DIFF
--- a/src/modules/socket/client/api_client.cpp
+++ b/src/modules/socket/client/api_client.cpp
@@ -11,6 +11,7 @@
 #include "socket/api.hpp"
 #include "socket/process_control.hpp"
 #include "socket/client/api_client.hpp"
+#include "socket/client/channels_hashtab.hpp"
 
 namespace openperf::socket::api {
 
--- a/src/modules/socket/client/api_client.hpp
+++ b/src/modules/socket/client/api_client.hpp
@@ -13,8 +13,6 @@
 #include "socket/unix_socket.hpp"
 #include "core/op_uuid.hpp"
 
-#include "channels_hashtab.hpp"
-
 namespace openperf::socket::api {
 
 template <typename T> class thread_singleton
--- a/src/modules/socket/client/channels_hashtab.hpp
+++ b/src/modules/socket/client/channels_hashtab.hpp
@@ -1,6 +1,8 @@
 #ifndef _OP_SOCKET_API_CHANNELS_HASHTAB_HPP_
 #define _OP_SOCKET_API_CHANNELS_HASHTAB_HPP_
 
+#include <shared_mutex>
+
 #include "core/op_hashtab.h"
 #include "socket/client/io_channel_wrapper.hpp"
 #include "utils/singleton.hpp"
@@ -33,6 +35,7 @@ public:
                        int client_fd,
                        int server_fd)
     {
+        std::shared_lock<std::shared_mutex> m_guard(m_lock);
         auto value = new ided_channel{
             id, io_channel_wrapper(channel, client_fd, server_fd)};
         return op_hashtab_insert(
@@ -41,17 +44,35 @@ public:
 
     inline ided_channel* find(int fd)
     {
+        std::shared_lock<std::shared_mutex> m_guard(m_lock);
         return reinterpret_cast<ided_channel*>(
             op_hashtab_find(m_tab, reinterpret_cast<void*>(fd)));
     };
 
     inline bool erase(int fd)
     {
-        return op_hashtab_delete(m_tab, reinterpret_cast<void*>(fd));
+        // hashtab garbage collection requires exclusive lock
+        std::lock_guard<std::shared_mutex> m_guard(m_lock);
+
+        auto idc = reinterpret_cast<ided_channel*>(
+            op_hashtab_find(m_tab, reinterpret_cast<void*>(fd)));
+        if (!idc) return false;
+
+        // Mark the channel deleted
+        op_hashtab_delete(m_tab, reinterpret_cast<void*>(fd));
+
+        // Perform garbage collection to actually delete the channel object
+        op_hashtab_garbage_collect(m_tab);
+        return true;
     };
 
 private:
     op_hashtab* m_tab;
+
+    // This shared mutex is used only used to obtain exclusive access
+    // when doing garbage collection.  All other hashtab operations
+    // are lock free thread safe so they don't need exclusive lock.
+    std::shared_mutex m_lock;
 };
 } // namespace openperf::socket::api
 
--- a/src/modules/socket/client/channels_hashtab.hpp
+++ b/src/modules/socket/client/channels_hashtab.hpp
@@ -16,13 +16,14 @@ class channels_hashtab : public utils::singleton<channels_hashtab>
         socket_id id;
         io_channel_wrapper channel;
     };
+    using ided_channel_ptr = std::shared_ptr<ided_channel>;
 
 public:
     channels_hashtab()
     {
         m_tab = op_hashtab_allocate();
         auto descructor = [](void* value) -> void {
-            auto channel = reinterpret_cast<ided_channel*>(value);
+            auto channel = reinterpret_cast<ided_channel_ptr*>(value);
             delete channel;
         };
         op_hashtab_set_value_destructor(m_tab, descructor);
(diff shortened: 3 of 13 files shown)


DerangedMonkeyNinja (changes requested):
> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/459)* status: all files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent)

___
*[src/modules/socket/client/channels_hashtab.hpp, line 80 at r1](https://reviewable.io/reviews/spirent/openperf/459#-MRu8Re83aslBnRMxv2v:-MRu8Re83aslBnRMxv2w:b44yuzd) ([raw file](https://github.com/spirent/openperf/blob/a5debf4bd953e6a88cca5f310337a0284ea878c7/src/modules/socket/client/channels_hashtab.hpp#L80)):*
> ```cpp
>         // Need to mark the channel deleted before calling close() function
>         // to avoid getting overridden close() when using LD_PRELOAD.
>         idc->channel.close();
> ```

The explicit close seems unnecessary...  Wouldn't calling `close()` in the destructor be sufficient?  What am I missing?

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/459)* status: all files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/socket/client/channels_hashtab.hpp, line 80 at r1](https://reviewable.io/reviews/spirent/openperf/459#-MRu8Re83aslBnRMxv2v:-MRvCbMh-CIYk9cIY8jX:bas0dr9) ([raw file](https://github.com/spirent/openperf/blob/a5debf4bd953e6a88cca5f310337a0284ea878c7/src/modules/socket/client/channels_hashtab.hpp#L80)):*

The explicit close seems unnecessary...  Wouldn't calling `close()` in the destructor be sufficient?  What am I missing?
</blockquote></details>

That would require gabage collect every time close() is called.
Currently garbage collection is only done on every 10th close.
Maybe premature optimization, but I was trying to reduce exclusive locking and number of times garbage collection is done.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/459)* status: all files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent)

___
*[src/modules/socket/client/channels_hashtab.hpp, line 80 at r1](https://reviewable.io/reviews/spirent/openperf/459#-MRu8Re83aslBnRMxv2v:-MRz96rJ9pIc8GqD75J7:b-95bi66) ([raw file](https://github.com/spirent/openperf/blob/a5debf4bd953e6a88cca5f310337a0284ea878c7/src/modules/socket/client/channels_hashtab.hpp#L80)):*

That would require gabage collect every time close() is called.
Currently garbage collection is only done on every 10th close.
Maybe premature optimization, but I was trying to reduce exclusive locking and number of times garbage collection is done.
</blockquote></details>

Not exactly.

The hashtab allows you to set destructor functions for both keys and values. Those destructors aren't called until the `garbage_collect` function is called.  The intent is to tell the hashtab how to destroy the objects it holds and it will destroy them when it is safe to do so.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/459)* status: all files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/socket/client/channels_hashtab.hpp, line 80 at r1](https://reviewable.io/reviews/spirent/openperf/459#-MRu8Re83aslBnRMxv2v:-MS-S3rB4IlE-AYscWS4:bmonx9b) ([raw file](https://github.com/spirent/openperf/blob/a5debf4bd953e6a88cca5f310337a0284ea878c7/src/modules/socket/client/channels_hashtab.hpp#L80)):*

Not exactly.

The hashtab allows you to set destructor functions for both keys and values. Those destructors aren't called until the `garbage_collect` function is called.  The intent is to tell the hashtab how to destroy the objects it holds and it will destroy them when it is safe to do so.
</blockquote></details>

Sorry.  I am not sure what you are asking me to change.

The eventfds need to be closed in the socket client API close() function.
It can not be deferred until garbage collection if garbage collection is batched.
If you defer fd close then the fd used in epoll or other socket API calls will not behave normally and you could get extra notifications or other weird issues due to an fd still being left open.

https://man7.org/linux/man-pages/man7/epoll.7.html
6.  Will closing a file descriptor cause it to be removed from all epoll interest lists?
           Yes, but ....

So I either need to remove batching garbage collection or keep it as is with close() prior to garbage collection.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/459)* status: all files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent)

___
*[src/modules/socket/client/channels_hashtab.hpp, line 80 at r1](https://reviewable.io/reviews/spirent/openperf/459#-MRu8Re83aslBnRMxv2v:-MS-WEWL2h1xEngOW1-Z:b-veg649) ([raw file](https://github.com/spirent/openperf/blob/a5debf4bd953e6a88cca5f310337a0284ea878c7/src/modules/socket/client/channels_hashtab.hpp#L80)):*

Sorry.  I am not sure what you are asking me to change.

The eventfds need to be closed in the socket client API close() function.
It can not be deferred until garbage collection if garbage collection is batched.
If you defer fd close then the fd used in epoll or other socket API calls will not behave normally and you could get extra notifications or other weird issues due to an fd still being left open.

https://man7.org/linux/man-pages/man7/epoll.7.html
6.  Will closing a file descriptor cause it to be removed from all epoll interest lists?
           Yes, but ....

So I either need to remove batching garbage collection or keep it as is with close() prior to garbage collection.
</blockquote></details>

Ahh, you're right...

I think you should remove the batched garbage collection. When a channel is removed from the hashtab, it's still potentially accessible to other threads.  Any query that happens concurrently with the delete may or may not find the deleted item. That's why the hashtab doesn't actually delete anything unless `garbage_collect()` is explicitly called.  Since you're closing the fd's at this point, we need to make sure no other threads can retrieve this channel, so we need to grab the exclusive lock at this point anyway.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/459)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/modules/socket/client/channels_hashtab.hpp, line 80 at r1](https://reviewable.io/reviews/spirent/openperf/459#-MRu8Re83aslBnRMxv2v:-MS7AX0qEi7fgtmxGiro:bqe81ce) ([raw file](https://github.com/spirent/openperf/blob/a5debf4bd953e6a88cca5f310337a0284ea878c7/src/modules/socket/client/channels_hashtab.hpp#L80)):*

Ahh, you're right...

I think you should remove the batched garbage collection. When a channel is removed from the hashtab, it's still potentially accessible to other threads.  Any query that happens concurrently with the delete may or may not find the deleted item. That's why the hashtab doesn't actually delete anything unless `garbage_collect()` is explicitly called.  Since you're closing the fd's at this point, we need to make sure no other threads can retrieve this channel, so we need to grab the exclusive lock at this point anyway.
</blockquote></details>

I had to move the close() outside of the destructor even without batched garbage collect because it was causing issues when using it with LD_PRELOAD.  When close() is overridden the overridden close() was getting called again in the destructor and since shared_mutex it is not reentrant it was getting stuck.  Without switching to recursive_mutex or implementing a recusive_shared_mutex we have to do close() without holding the exclusive lock.

DerangedMonkeyNinja (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/459)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cmyers-spirent)

DerangedMonkeyNinja (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/459#-:-MS8hTkb67con1CR3N72:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cmyers-spirent)

PR #460, opened 2021-01-27
Packet Generator ND Support

dmorton-spirent:
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/460)


DerangedMonkeyNinja (changes requested):
> Reviewed 13 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/460)* status: 13 of 24 files reviewed, 4 unresolved discussions (waiting on @dmorton-spirent)

___
*[src/modules/packet/generator/interface_source.cpp, line 63 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3Oy_026zagmJBlwX1:-MS3Oy_1Aj5notYPMlZT:b-td93kd) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/interface_source.cpp#L63)):*
> ```cpp
>     std::tuple<libpacket::type::ipv6_address, // Interface Global Address
>                libpacket::type::ipv6_address, // Interface Link-Local Address
>                bool // true if three address above were found and valid. false
> ```

three? Would it be simpler to use a tuple of optionals?  Or maybe an optional tuple? You know how I feel about context-free booleans...
___
*[src/modules/packet/generator/interface_source.cpp, line 183 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3OTByCy7JNGaPG0k1:-MS3OTByCy7JNGaPG0k2:bciz9ak) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/interface_source.cpp#L183)):*
> ```cpp
>     if (std::holds_alternative<unresolved>(
>             maybe_result_entry->second.next_hop_mac)) {
>         OP_LOG(OP_LOG_ERROR,
> ```

Seems more like a WARNing?
___
*[src/modules/packet/generator/learning.hpp, line 30 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3PLYACB4Gviemlo7q:-MS3PLYACB4Gviemlo7r:b-6w25by) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/learning.hpp#L30)):*
> ```cpp
>     std::unordered_map<libpacket::type::ipv4_address, mac_type>;
> 
> using ipv6_type = std::variant<unresolved, libpacket::type::ipv6_address>;
> ```

I didn't notice this before...  Why use a two state variant here instead of an optional?
___
*[src/modules/packet/generator/learning.cpp, line 19 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3R-M25JeW3sKmaBhl:-MS3R-M25JeW3sKmaBhm:b-a9ogez) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/learning.cpp#L19)):*
> ```cpp
> static constexpr int max_poll_count = 30;
> 
> static constexpr int nd_neighbor_cache_no_entry =
> ```

Suggest using the same format as other lwip error conditions, e.g. ERR_NOT_FOUND or something.
Also, please use a unique value. -10 is already used by ERR_ISCONN. Since lwip uses a signed 8 bit value for err_t, I'd suggest starting at the minimum value and working up for yours, so -127. That minimizes any chance of having a collision with these error codes.

dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/460)* status: 13 of 24 files reviewed, 4 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/packet/generator/interface_source.cpp, line 63 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3Oy_026zagmJBlwX1:-MS4Xxfx23ulagM8KO-C:b-rafd93) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/interface_source.cpp#L63)):*

three? Would it be simpler to use a tuple of optionals?  Or maybe an optional tuple? You know how I feel about context-free booleans...
</blockquote></details>

Good point. Context-free booleans are a tough habit to break :).

I'll make it a tuple of `std::optionals` and update the population logic. Thinking about it I realized there are valid cases for interfaces with only a link-local address. Will make the same change for IPv4 since you can have a perfectly valid configuration without a gateway address.
___
*[src/modules/packet/generator/interface_source.cpp, line 183 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3OTByCy7JNGaPG0k1:-MS4Y_t860cQEL8T-v6V:bin9coh) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/interface_source.cpp#L183)):*

Seems more like a WARNing?
</blockquote></details>

Will downgrade to a warning; no strong feeling either way here.
___
*[src/modules/packet/generator/learning.hpp, line 30 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3PLYACB4Gviemlo7q:-MS4ZXkREeQaEcb_vAR_:bvjrzwg) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/learning.hpp#L30)):*

I didn't notice this before...  Why use a two state variant here instead of an optional?
</blockquote></details>

I was in state-as-variant mode because I was writing the state machine at the same time. I can make the same arguments for readability for both. Unless `std::holds_alternative` and `std::visit` and `std::get` have significantly worse performance than `std::optional<T>.has_value()`?
___
*[src/modules/packet/generator/learning.cpp, line 19 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3R-M25JeW3sKmaBhl:-MS4YkBLD8DmbEl59d7C:b-37fn1q) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/learning.cpp#L19)):*

Suggest using the same format as other lwip error conditions, e.g. ERR_NOT_FOUND or something.
Also, please use a unique value. -10 is already used by ERR_ISCONN. Since lwip uses a signed 8 bit value for err_t, I'd suggest starting at the minimum value and working up for yours, so -127. That minimizes any chance of having a collision with these error codes.
</blockquote></details>

Understood. Will change.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/460)* status: 13 of 24 files reviewed, 2 unresolved discussions (waiting on @dmorton-spirent)

___
*[src/modules/packet/generator/interface_source.cpp, line 183 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3OTByCy7JNGaPG0k1:-MS4otxu9y6XG9sobH8O:b-9i0fwc) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/interface_source.cpp#L183)):*

Will downgrade to a warning; no strong feeling either way here.
</blockquote></details>

Generally, reserve ERROR for when the system or user did something wrong.  If it's just a normal, but unexpected event, then WARN is more appropriate.
___
*[src/modules/packet/generator/learning.hpp, line 30 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3PLYACB4Gviemlo7q:-MS4rC6v7tmJBNtXctn1:b7c3x5l) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/learning.hpp#L30)):*

I was in state-as-variant mode because I was writing the state machine at the same time. I can make the same arguments for readability for both. Unless `std::holds_alternative` and `std::visit` and `std::get` have significantly worse performance than `std::optional<T>.has_value()`?
</blockquote></details>

You wouldn't have those no-op variant functions to write, so the code would be cleaner. I doubt it matters performance wise. Also, `std::optional<T>` is implicitly converted to bool.  You can write `if (my_optional) {}` instead of `if (my_optional.has_value()) {}`.

dmorton-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/460)* status: 9 of 24 files reviewed, 2 unresolved discussions (waiting on @DerangedMonkeyNinja)

___
*[src/modules/packet/generator/interface_source.cpp, line 183 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3OTByCy7JNGaPG0k1:-MS5xaKM8VxQdK837eej:b8jk320) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/interface_source.cpp#L183)):*

Generally, reserve ERROR for when the system or user did something wrong.  If it's just a normal, but unexpected event, then WARN is more appropriate.
</blockquote></details>

Understood; thanks for the info.
___
*[src/modules/packet/generator/learning.hpp, line 30 at r1](https://reviewable.io/reviews/spirent/openperf/460#-MS3PLYACB4Gviemlo7q:-MS51_aV08pb3rArQJWG:bv0kcs8) ([raw file](https://github.com/spirent/openperf/blob/5bd4d41e3098126a372d508306a641dc44473211/src/modules/packet/generator/learning.hpp#L30)):*

You wouldn't have those no-op variant functions to write, so the code would be cleaner. I doubt it matters performance wise. Also, `std::optional<T>` is implicitly converted to bool.  You can write `if (my_optional) {}` instead of `if (my_optional.has_value()) {}`.
</blockquote></details>

Got it. Changed over.

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 1 of 24 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/460#-:-MS8iam25mtsgqR8Wl4P:bnfp4nl)* status: 14 of 24 files reviewed, all discussions resolved

PR #464, opened 2021-02-19
Socket api fixes

cmyers-spirent:
Misc framework, lwip, socket api issues found when testing the network load generator

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/464)


src/framework/core/op_list.c
```diff
@@ -175,6 +185,7 @@ int op_list_purge(struct op_list* list)
  */
 void op_list_garbage_collect(struct op_list* list)
 {
+    _op_list_collect_deleted(list);
```
cmyers-spirent, line 188: This was causing issues with channels_hashtab because some of the channel objects were not getting deleted when op_hashtab_garbage_collect() was called.

The channel shared memory was getting reused by the next garbage collection call causing very strange issues with the different fds from expected getting closed when the channel object was deleted.

DerangedMonkeyNinja (approved):
:lgtm:

Just one question though...  Should we merge `netifapi_netif_safe_set_down()` upstream? Or are we doing something weird with the stack that requires it?

> Reviewed 4 of 5 files at r1, 11 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/464#-:-MTuaeQ-8C8EnuQiawHF:b-cqnrro)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cmyers-spirent)

cmyers-spirent (commented):
It is kind of a corner case, so not sure how it should be handled in lwip generically.
If you call netif_remove() with the interface up you should get this behavior.

Issues occur because we set the interface down before removal which prevents proper tcp abort/rst from going out and arp holds  from getting cleared.  Should we change openperf code to do that?
It would make the net_interface code a bit more confusing.

Main issue would be that net_interface::down() could cause intermittent crashes if not used properly in future changes , so I thought it was safer to fix it so down() was always safe.  I guess it would be safer if lwip cleared the arp hold queue even if the interface is down so it won't crashes, but tcp sessions don't get terminated so it is not a polite way to remove the interface and people should do that...

I guess in real world it is still desirable to leave tcp sessions up when interface is down so you could reset hardware etc and bring interface up with sessions still working.  I don't think openperf has need to support that though and more import to have tcp rsts go out.

> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/464#-:-MTvk0nV1aoUngN6E5UG:b-1q1p3k)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cmyers-spirent)

PR #467, opened 2021-02-26
Network fixes

cmyers-spirent:
Sorry lots of changes.   The main changes are:
- Change all busy rx loops to use poll/epoll to prevent wasting cpu
- Added network generator "toggle" operations to switch load generation from running load generator to a non-running load generator

The toggle is done by swapping the controller object between the running network generator and the idle network generator and updating the stats (via generator pointer) and sync object pointers to the idle load generator.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/467)


DerangedMonkeyNinja (changes requested):
> Reviewed 28 of 43 files at r1, 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 43 files reviewed, 7 unresolved discussions (waiting on @cmyers-spirent)

___
*[src/framework/generator/controller.hpp, line 182 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT6GtP8acHNLSea_hf:-MUT6GtP8acHNLSea_hg:buwk7rz) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/controller.hpp#L182)):*
> ```cpp
>     std::vector<T*> tasks;
>     for (auto& worker : m_workers) {
>         auto task = dynamic_cast<T*>(worker.get_task());
> ```

Surely you mean static cast here?  The caller is providing the type, after all.
___
*[src/framework/generator/controller.hpp, line 192 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT6UKO8lPc_g9YomP9:-MUT6UKO8lPc_g9YomPA:b-1kzc9c) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/controller.hpp#L192)):*
> ```cpp
> {
>     m_workers.remove_if([this, task](auto& worker) {
>         auto worker_task = dynamic_cast<T*>(worker.get_task());
> ```

And here.
___
*[src/framework/generator/worker.hpp, line 81 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTIEZv8dlxo1--izZM:-MUTIEZv8dlxo1--izZN:btqgpzz) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/worker.hpp#L81)):*
> ```cpp
>     // m_task stores the base class so pass derived class in task_ptr
>     // to the run() function because that is also a template function.
>     T* task_ptr = new T(std::move(task));
> ```

Yuck. I guess this is the source of the dynamic cast.
Why do you need a new, heap allocated object here?
___
*[src/modules/cpu/server.cpp, line 2 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT79wWAA-yJFB0EM69:-MUT79wWAA-yJFB0EM6A:b-krvdd6) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/cpu/server.cpp#L2)):*
> ```cpp
> #include <zmq.h>
> #include <forward_list>
> ```

Having to add this to a compilation unit means some header needed it, but didn't add it right?
___
*[src/modules/network/generator_stack.hpp, line 49 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT8XBd1ae9YF13wMyr:-MUT8XBd1ae9YF13wMys:b-nwoiwm) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/generator_stack.hpp#L49)):*
> ```cpp
> 
>     tl::expected<model::generator_result, std::string>
>     toggle_generator(const std::string& out_id,
> ```

std::string_view?
___
*[src/modules/network/task.cpp, line 533 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT9ts23hMM9yCiOqqf:-MUT9ts23hMM9yCiOqqg:b-3lzxpz) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/task.cpp#L533)):*
> ```cpp
>     ssize_t recv_or_err = 0;
> 
>     if (m_spin_ops_left == 0) {
> ```

Do these additions mean there are state issues with the connections?
___
*[src/modules/network/firehose/server_tcp.cpp, line 226 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTGY0L5Bfnvi6XBGsz:-MUTGY0MDTfA_Kh3hoJV:b-dkbsih) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/firehose/server_tcp.cpp#L226)):*
> ```cpp
>         if (npoll <= 0) { continue; }
> 
>         if (pfd[0].revents & POLLIN) {
> ```

What's the point of the eventfd here?

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 43 files reviewed, 7 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/framework/generator/controller.hpp, line 4 at r1](https://reviewable.io/reviews/spirent/openperf/467#-MUQxTSr7zEqC7eY9gSI:-MUQxTSr7zEqC7eY9gSJ:bza3q94) ([raw file](https://github.com/spirent/openperf/blob/bbc875b7f2bde3018027e71a072042f1b9737c02/src/framework/generator/controller.hpp#L4)):*
> ```cpp
> #define _OP_FRAMEWORK_GENERATOR_CONTROLLER_HPP_
> 
> #include <cassert>
> ```

The optimization of using a forward_list instead of list and reversing the list due to no emplace_back() seemed kind of unnecessary.
So changed it to a list instead so order is preserved.
___
*[src/framework/generator/controller.hpp, line 182 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT6GtP8acHNLSea_hf:-MUoc0MA8LKsJdPrXgq7:b-tkqxcd) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/controller.hpp#L182)):*

Surely you mean static cast here?  The caller is providing the type, after all.
</blockquote></details>

add() is a template function, so technically it allows adding different classes to the same controller.
So to keep this consistent with add() we need dynamic cast otherwise we need to return base class and require calling code to cast.
As far as I can tell it looks like each controller only handles a single class,  though.
___
*[src/framework/generator/task.hpp, line 5 at r1](https://reviewable.io/reviews/spirent/openperf/467#-MUQy6Fn1Qa9JsnDbPzw:-MUQy6Fn1Qa9JsnDbPzx:bq1mt6e) ([raw file](https://github.com/spirent/openperf/blob/bbc875b7f2bde3018027e71a072042f1b9737c02/src/framework/generator/task.hpp#L5)):*
> ```cpp
> 
> namespace openperf::framework::generator {
> ```

Added virtual base class so the controller can do dynamic_cast
___
*[src/framework/generator/worker.hpp, line 81 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTIEZv8dlxo1--izZM:-MUoeekl6i4LyaA7DLP5:b-6heq2w) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/worker.hpp#L81)):*

Yuck. I guess this is the source of the dynamic cast.
Why do you need a new, heap allocated object here?
</blockquote></details>

Need access to the task pointer so task can be reconfigured.
If we don't alloc an object we need to wait until the started std::thread can set a member variable pointing to the task and clear pointer on exit.  otherwise we have race conditions.
Seemed better to alloc an object instead of adding a spin loop.

IMHO the task should be alloc by the caller and passed through unique_ptr, but then I have to change all the load generators.  I was trying to limit changes to just the network load generator...
___
*[src/modules/cpu/server.cpp, line 2 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT79wWAA-yJFB0EM69:-MUoffqK-lqzXCt6J9AF:bu5mo3m) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/cpu/server.cpp#L2)):*

Having to add this to a compilation unit means some header needed it, but didn't add it right?
</blockquote></details>

correct. the controller was using forward_list which was changed to list.
___
*[src/modules/network/task.cpp, line 533 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT9ts23hMM9yCiOqqf:-MUog2LK3XCOVASf5rH-:bxlvccc) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/task.cpp#L533)):*

Do these additions mean there are state issues with the connections?
</blockquote></details>

Original code was handling all connections in a socket dispatch loop without callbacks, so it didn't need as much state because the loop itself was keeping track of when it was done.
Since we need to handle connections through callbacks, additional state  needed to be added to the class so that callbacks functions can know when to tell the event loop dispatch to exit.
___
*[src/modules/network/firehose/server_tcp.cpp, line 226 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTGY0L5Bfnvi6XBGsz:-MUohImaAfAehiZGlbRL:b-plc5tq) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/firehose/server_tcp.cpp#L226)):*

What's the point of the eventfd here? 
</blockquote></details>

It allows you to wakeup the accept thread.  otherwise, the thread will block on accept().

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 43 files reviewed, 7 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/generator_stack.hpp, line 49 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT8XBd1ae9YF13wMyr:-MUoo_4TDOCJc8PdiMkC:b-mp5jv6) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/generator_stack.hpp#L49)):*

std::string_view?
</blockquote></details>

It is consistent with other functions in the same class.
I think functions were written with std::string because there are 2 unordered_maps with std::string keys.
If member function does a map find()/at()/erase() it needs to construct a std::string from the std::string_view.
So could change it to std::string_view, but would need to change all functions and 
have functions to construct strings e.g.:
auto out_gen = m_generators.at(std::string(out_id));

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 43 files reviewed, 7 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/network/generator_stack.hpp, line 49 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT8XBd1ae9YF13wMyr:-MUoqmUkEwds58tfMCpX:bk979xy) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/generator_stack.hpp#L49)):*

It is consistent with other functions in the same class.
I think functions were written with std::string because there are 2 unordered_maps with std::string keys.
If member function does a map find()/at()/erase() it needs to construct a std::string from the std::string_view.
So could change it to std::string_view, but would need to change all functions and 
have functions to construct strings e.g.:
auto out_gen = m_generators.at(std::string(out_id));
</blockquote></details>

So doesn't seem like it makes sense to change it.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 43 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent)

___
*[src/framework/generator/controller.hpp, line 182 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT6GtP8acHNLSea_hf:-MUsM-ZTB9H7ZLIY6yK5:b-ki1arf) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/controller.hpp#L182)):*

add() is a template function, so technically it allows adding different classes to the same controller.
So to keep this consistent with add() we need dynamic cast otherwise we need to return base class and require calling code to cast.
As far as I can tell it looks like each controller only handles a single class,  though.
</blockquote></details>

Yeah, I'm pretty sure the controllers only handle a single class.  I'm not following the comment of `add()`, though.  It doesn't have a dynamic_cast...

Anyway, please remove the dynamic_casts here. If client code has a type problem, that needs to be fixed in the clients, not here.
___
*[src/framework/generator/worker.hpp, line 81 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTIEZv8dlxo1--izZM:-MUsNPm38GF619a7wVVA:b-6isub9) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/worker.hpp#L81)):*

Need access to the task pointer so task can be reconfigured.
If we don't alloc an object we need to wait until the started std::thread can set a member variable pointing to the task and clear pointer on exit.  otherwise we have race conditions.
Seemed better to alloc an object instead of adding a spin loop.

IMHO the task should be alloc by the caller and passed through unique_ptr, but then I have to change all the load generators.  I was trying to limit changes to just the network load generator...
</blockquote></details>

Please clean up problems as you find them!  If an improvement requires changing all clients, so be it.  Use it as an opportunity to clean up and simplify.
___
*[src/modules/network/task.cpp, line 533 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT9ts23hMM9yCiOqqf:-MUsNp9oA4L7i7Vya2Y4:b-7jg2oe) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/task.cpp#L533)):*

Original code was handling all connections in a socket dispatch loop without callbacks, so it didn't need as much state because the loop itself was keeping track of when it was done.
Since we need to handle connections through callbacks, additional state  needed to be added to the class so that callbacks functions can know when to tell the event loop dispatch to exit.
</blockquote></details>

sigh.
___
*[src/modules/network/firehose/server_tcp.cpp, line 226 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTGY0L5Bfnvi6XBGsz:-MUsL1CYArAUcd1QyxRn:bkmtluh) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/firehose/server_tcp.cpp#L226)):*

It allows you to wakeup the accept thread.  otherwise, the thread will block on accept().
</blockquote></details>

But...  this branch doesn't seem to do anything except clear the fd.  Why poll for it if waking up doesn't cause this function to do anything?

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 43 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/framework/generator/controller.hpp, line 182 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT6GtP8acHNLSea_hf:-MUt_I_h9CVdDg-Jk-wF:b-9l2o08) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/controller.hpp#L182)):*

Yeah, I'm pretty sure the controllers only handle a single class.  I'm not following the comment of `add()`, though.  It doesn't have a dynamic_cast...

Anyway, please remove the dynamic_casts here. If client code has a type problem, that needs to be fixed in the clients, not here.
</blockquote></details>

It is not a client issue, it is a controller class design issue to prevent improper usage.

add() doesn't have dynamic_cast but it does allow adding different classes to the same controller, so you either need to use dynamic_cast to pull out the ones you want or return task_base class and have users dynamic_cast<>.

class A : task<A> { ... }
class B : task<B> { ...}
controller.add(A(), a_name);
controller.add(B(), b_name);
auto tasks = controller.get_tasks<A>();
without dynamic_cast get_tasks() will return A and B.

So get_task() either needs dynamic_cast, needs to just return base class and have caller dynamic_cast, or controller needs to be changed so it only supports a single class type.  otherwise, the controller class can be used incorrectly which is bad design.
___
*[src/framework/generator/worker.hpp, line 81 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTIEZv8dlxo1--izZM:-MUteKRz42Vt1NM4VEBm:bkprn34) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/worker.hpp#L81)):*

Please clean up problems as you find them!  If an improvement requires changing all clients, so be it.  Use it as an opportunity to clean up and simplify.
</blockquote></details>

ok.  I will change it so that the task is allocated by the caller and passed through unique_ptr<>
___
*[src/modules/network/firehose/server_tcp.cpp, line 226 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTGY0L5Bfnvi6XBGsz:-MUtbGn9Bt8dC-12fLuU:b-aqz3zh) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/firehose/server_tcp.cpp#L226)):*

But...  this branch doesn't seem to do anything except clear the fd.  Why poll for it if waking up doesn't cause this function to do anything?
</blockquote></details>

Can't use accept() directly because that will block forever.
Original code would just close the accept fd from another thread to get it to exit, but that could cause race conditions with the DPDK socket implementation.
So I added a poll() which checks the accept_fd and the event_fd so we can sleep and still get woken up.
The stop() function toggles a m_running flag and wakes up the accept thread using the eventfd.  It is read here just to be nice because we are going to exit the loop anyway due to m_running flag getting toggled.
It seemed like a lot of unnecessary complexity to add a zmq and event_loop just to handle the exit case and event_fd+poll provides an easy way to prevent blocking on accept().

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 43 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/framework/generator/controller.hpp, line 182 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT6GtP8acHNLSea_hf:-MUtfZlQCyMvuOVQlOhU:blny9df) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/controller.hpp#L182)):*

It is not a client issue, it is a controller class design issue to prevent improper usage.

add() doesn't have dynamic_cast but it does allow adding different classes to the same controller, so you either need to use dynamic_cast to pull out the ones you want or return task_base class and have users dynamic_cast<>.

class A : task<A> { ... }
class B : task<B> { ...}
controller.add(A(), a_name);
controller.add(B(), b_name);
auto tasks = controller.get_tasks<A>();
without dynamic_cast get_tasks() will return A and B.

So get_task() either needs dynamic_cast, needs to just return base class and have caller dynamic_cast, or controller needs to be changed so it only supports a single class type.  otherwise, the controller class can be used incorrectly which is bad design.
</blockquote></details>

not sure why it turned bold, that was not intentional

DerangedMonkeyNinja (changes requested):
> Reviewed 6 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 46 files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/framework/generator/controller.hpp, line 182 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT6GtP8acHNLSea_hf:-MUxVnsw03j9jbiVCa0N:b-xnjb8e) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/controller.hpp#L182)):*

not sure why it turned bold, that was not intentional
</blockquote></details>

If you're worried about that, add static asserts to explicitly verify types.
In that case, clients can't misuse it and will know at compile time.
___
*[src/modules/network/firehose/server_tcp.cpp, line 226 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTGY0L5Bfnvi6XBGsz:-MUxVKbc-FNV9l_vRmUl:bzbgin0) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/firehose/server_tcp.cpp#L226)):*

Can't use accept() directly because that will block forever.
Original code would just close the accept fd from another thread to get it to exit, but that could cause race conditions with the DPDK socket implementation.
So I added a poll() which checks the accept_fd and the event_fd so we can sleep and still get woken up.
The stop() function toggles a m_running flag and wakes up the accept thread using the eventfd.  It is read here just to be nice because we are going to exit the loop anyway due to m_running flag getting toggled.
It seemed like a lot of unnecessary complexity to add a zmq and event_loop just to handle the exit case and event_fd+poll provides an easy way to prevent blocking on accept().
</blockquote></details>

Ok.  Could you add a comment here to explain that?  It's totally non-obvious and I want the next person who comes along to have some inclination of what's happening.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 33 of 46 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/framework/generator/controller.hpp, line 182 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUT6GtP8acHNLSea_hf:-MUyWyhU2HkTFkwJDazF:b-6j0qbb) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/framework/generator/controller.hpp#L182)):*

If you're worried about that, add static asserts to explicitly verify types.
In that case, clients can't misuse it and will know at compile time.
</blockquote></details>

huh? how would you do that with a static assert?
in any case, it is not an issue anymore because I changed get_tasks() and remove_task() to use task_base.
___
*[src/modules/network/firehose/server_tcp.cpp, line 226 at r2](https://reviewable.io/reviews/spirent/openperf/467#-MUTGY0L5Bfnvi6XBGsz:-MUyYssIEHh_ZuNoqIX2:b32wtd) ([raw file](https://github.com/spirent/openperf/blob/ad3ef80450624878be2d762ac9303a991bdd618d/src/modules/network/firehose/server_tcp.cpp#L226)):*

Ok.  Could you add a comment here to explain that?  It's totally non-obvious and I want the next person who comes along to have some inclination of what's happening.
</blockquote></details>

ok

DerangedMonkeyNinja (commented):
> Reviewed 3 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467)* status: 36 of 46 files reviewed, all discussions resolved

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/467#-:-MUyp1w353sYze5kJxoN:bnfp4nl)* status: 36 of 46 files reviewed, all discussions resolved

PR #475, opened 2021-03-17
framework: op_event_loop ignore epoll_wait() EINTR

cmyers-spirent:
This was causing all threads using op_event_loop to exit
when breaking and continuing with gdb.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/475)

### DIFF
--- a/src/framework/core/op_event_loop_epoll.c
+++ b/src/framework/core/op_event_loop_epoll.c
@@ -664,9 +664,11 @@ int op_event_loop_run_timeout(struct op_event_loop* loop, int timeout)
                 _do_event_handling(loop, events, nb_events);
                 _do_event_removals(loop);
             } else if (nb_events < 0) {
-                /* epoll error */
-                op_event_loop_exit(loop);
-                return (-errno);
+                if (errno != EINTR) {
+                    /* epoll error */
+                    op_event_loop_exit(loop);
+                    return (-errno);
+                }
             } else if (SLIST_EMPTY(&loop->always_list)) {
                 /* no ready epoll events and empty always list; must be a
                  * timeout */


DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/475)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent)

___
*[src/framework/core/op_event_loop_epoll.c, line 667 at r1](https://reviewable.io/reviews/spirent/openperf/475#-MW-NqWJ61ia93-0bskl:-MW-NqWJ61ia93-0bskm:b3m59ys) ([raw file](https://github.com/spirent/openperf/blob/6301e11a8d02b41411cacc4a81bc5387053486a3/src/framework/core/op_event_loop_epoll.c#L667)):*
> ```C
>                 _do_event_removals(loop);
>             } else if (nb_events < 0) {
>                 if (errno != EINTR) {
> ```

Isn't EINTR also reported for Ctrl-C?  We definitely need the loop to exit for that case...

DerangedMonkeyNinja (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/475)* status: all files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent)

___
*[src/framework/core/op_event_loop_epoll.c, line 667 at r1](https://reviewable.io/reviews/spirent/openperf/475#-MW-NqWJ61ia93-0bskl:-MW-QKN-6stq9LmVemDr:bg5fysm) ([raw file](https://github.com/spirent/openperf/blob/6301e11a8d02b41411cacc4a81bc5387053486a3/src/framework/core/op_event_loop_epoll.c#L667)):*

Isn't EINTR also reported for Ctrl-C?  We definitely need the loop to exit for that case...
</blockquote></details>

There is currently a sighandler in main.c that catches SIGTERM/SIGINT.  I guess a proper fix for that would be adding some framework code to track the actual program state. The `main` functions could then use the framework functions to set the state and the loops could use another function to check if they should exit or not.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/475)* status: all files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

___
*[src/framework/core/op_event_loop_epoll.c, line 667 at r1](https://reviewable.io/reviews/spirent/openperf/475#-MW-NqWJ61ia93-0bskl:-MW0LGhRChnFKM8tm7UH:bow8rkq) ([raw file](https://github.com/spirent/openperf/blob/6301e11a8d02b41411cacc4a81bc5387053486a3/src/framework/core/op_event_loop_epoll.c#L667)):*

There is currently a sighandler in main.c that catches SIGTERM/SIGINT.  I guess a proper fix for that would be adding some framework code to track the actual program state. The `main` functions could then use the framework functions to set the state and the loops could use another function to check if they should exit or not.
</blockquote></details>

If we are relying on EINTR to shutdown the event op that would be a problem because signals aren't guaranteed to go to each thread and the threads might not be in epoll_wait() when signal arrives.

I think all modules should be using a shutdown canary to catch zmq shutdown so when op_halt() closes down zmq they should exit.
https://github.com/Spirent/openperf/search?q=canary

The main problem this causes though is that shutting down zmq also kills the logging so that s probably we don't get any log messages for threads shutting down.  Which makes it a little difficult to tell if they are getting terminated correctly...

Maybe using a zmq pub/sub socket would be better, so threads can subscribe for shutdown notification ahead of the zmq_shutdown, but then still need some wait loop to wait for module to shutdown, so maybe better to add a stop function to the module and module state (running, stopping, stopped, etc.) so we can check to make sure all modules are stopped before killing zmq and logging.

But EINTR fix is separate from that.

DerangedMonkeyNinja (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/475#-:-MW0vVEV8JysHNP5Gj1k:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cmyers-spirent)

___
*[src/framework/core/op_event_loop_epoll.c, line 667 at r1](https://reviewable.io/reviews/spirent/openperf/475#-MW-NqWJ61ia93-0bskl:-MW0vEr48i06kpyEpv1V:bnnevnn) ([raw file](https://github.com/spirent/openperf/blob/6301e11a8d02b41411cacc4a81bc5387053486a3/src/framework/core/op_event_loop_epoll.c#L667)):*

If we are relying on EINTR to shutdown the event op that would be a problem because signals aren't guaranteed to go to each thread and the threads might not be in epoll_wait() when signal arrives.

I think all modules should be using a shutdown canary to catch zmq shutdown so when op_halt() closes down zmq they should exit.
https://github.com/Spirent/openperf/search?q=canary

The main problem this causes though is that shutting down zmq also kills the logging so that s probably we don't get any log messages for threads shutting down.  Which makes it a little difficult to tell if they are getting terminated correctly...

Maybe using a zmq pub/sub socket would be better, so threads can subscribe for shutdown notification ahead of the zmq_shutdown, but then still need some wait loop to wait for module to shutdown, so maybe better to add a stop function to the module and module state (running, stopping, stopped, etc.) so we can check to make sure all modules are stopped before killing zmq and logging.

But EINTR fix is separate from that.
</blockquote></details>

We're not relying on EINTR for shutdown. I must be misremembering past adventures with `epoll`.  So long as Ctrl-C still works, I'm fine with it.

PR #477, opened 2021-03-31
modules/socket: add TCP/UDP PCB socket stats

cmyers-spirent:
This change adds a REST API route /socket-stats for TCP/UDP PCB
socket stats.  It is useful when debugging LWIP TCP/UDP socket issues.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/477)


DerangedMonkeyNinja (changes requested):
> Reviewed 19 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/477)* status: 19 of 27 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent)

___
*[api/schema/v1/modules/socket.yaml, line 2 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7WenB0PDrRsYJ-rUw:-MX7WenB0PDrRsYJ-rUx:b-aueszh) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/api/schema/v1/modules/socket.yaml#L2)):*
> ```YAML
> paths:
>   /socket-stats:
> ```

This API is unlike any of the others.  For consistency, there needs to be a top level `/socket` path that returns all the socket objects and then individual `/socket/{id}` paths that let you query individual socket statistics directly. This also allows room to expand the `/socket` path in the future.
___
*[src/modules/socket/server/pcb_utils.hpp, line 77 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7S0981o58z5QtwqFU:-MX7S0981o58z5QtwqFV:b-lgkes1) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/src/modules/socket/server/pcb_utils.hpp#L77)):*
> ```cpp
> udp_pcb_stats get_udp_pcb_stats(const udp_pcb*);
> 
> std::list<socket_pcb_stats> get_all_socket_pcb_stats();
> ```

Since it looks like this function is just returning a block of stats to the caller, prefer using a vector over a list here.  Lists are poor choices for moving data unless you have some good reason for choosing it.
___
*[src/modules/socket/server/pcb_utils.cpp, line 58 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7TVEe3cu75RcEHuso:-MX7TVEe3cu75RcEHusp:b-3vx4xf) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/src/modules/socket/server/pcb_utils.cpp#L58)):*
> ```cpp
> }
> 
> static int get_addr_length(uint8_t af)
> ```

Since length can never be negative, prefer an unsigned type here.
___
*[src/modules/socket/server/pcb_utils.cpp, line 116 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7Vn1x69QrgnRhi43F:-MX7Vn1y9_FFo7L_jNya:b-3u2vj0) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/src/modules/socket/server/pcb_utils.cpp#L116)):*
> ```cpp
> }
> 
> std::list<socket_pcb_stats> get_all_socket_pcb_stats()
> ```

Ok... I see why this function returns a list.

I'd prefer this function just use the map to generate the set of stats and then transform the data into a vector at then end.  I don't want to set any precedents for using lists to return data.

djoyner (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/477)* status: 19 of 27 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/socket.yaml, line 2 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7WenB0PDrRsYJ-rUw:-MX7XoZ50iRBFOwYYsX9:btj5xic) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/api/schema/v1/modules/socket.yaml#L2)):*

This API is unlike any of the others.  For consistency, there needs to be a top level `/socket` path that returns all the socket objects and then individual `/socket/{id}` paths that let you query individual socket statistics directly. This also allows room to expand the `/socket` path in the future.
</blockquote></details>

Agree, but the top-level path should be `/sockets` (plural) since it is a collection resource.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/477)* status: 19 of 27 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/socket.yaml, line 2 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7WenB0PDrRsYJ-rUw:-MX8xlAb1pxLkD4MeH3A:b-xyvunw) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/api/schema/v1/modules/socket.yaml#L2)):*
<details><summary><i>Previously, djoyner (David Joyner) wrote…</i></summary><blockquote>

Agree, but the top-level path should be `/sockets` (plural) since it is a collection resource.
</blockquote></details>

ok, but id need to be a little werid.  currently I am merging socket stats and pcb stats to a single view because it is useful to see those things merged.
there could be sockets without pcbs (unlikely, probably a bug or intermittent state if it happens) and pcbs without sockets (very likely).

so I could convert the pcb pointer address to a uuid and use 0 uuid for the socket without pcb case.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/477)* status: 19 of 27 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @djoyner)

___
*[api/schema/v1/modules/socket.yaml, line 2 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7WenB0PDrRsYJ-rUw:-MX8zoWa-89-NToA3jRl:b-msqs8) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/api/schema/v1/modules/socket.yaml#L2)):*

ok, but id need to be a little werid.  currently I am merging socket stats and pcb stats to a single view because it is useful to see those things merged.
there could be sockets without pcbs (unlikely, probably a bug or intermittent state if it happens) and pcbs without sockets (very likely).

so I could convert the pcb pointer address to a uuid and use 0 uuid for the socket without pcb case.
</blockquote></details>

The `id` needs to unambiguously identify the socket/pcb, so overloading 0 is not a good option. If we really shouldn't have sockets without pcbs, then just logging a warning about the condition would probably be sufficient, especially if that case shouldn't happen anyway.

djoyner (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/477)* status: 19 of 27 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/socket.yaml, line 2 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7WenB0PDrRsYJ-rUw:-MX9HS2p7I8j1Fjnagvh:b-i2fps5) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/api/schema/v1/modules/socket.yaml#L2)):*

The `id` needs to unambiguously identify the socket/pcb, so overloading 0 is not a good option. If we really shouldn't have sockets without pcbs, then just logging a warning about the condition would probably be sufficient, especially if that case shouldn't happen anyway.
</blockquote></details>

Since they are separate structures with (apparently) separate lifetimes and identifiers, would it make more sense to surface `/sockets` separately from `/pcbs`? I'd rather see us structure the REST API in a way that more accurately reflects internals than for perceived client convenience. Also keep in mind that a representation of a socket (`/socket/{id}`) could include an optional representation of the associated PCB.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/477)* status: 19 of 27 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[api/schema/v1/modules/socket.yaml, line 2 at r1](https://reviewable.io/reviews/spirent/openperf/477#-MX7WenB0PDrRsYJ-rUw:-MXA1TNgA55dmek39rCv:b-w9md91) ([raw file](https://github.com/spirent/openperf/blob/23dfa75129312b3791fe1f38ca32dd980d8ee898/api/schema/v1/modules/socket.yaml#L2)):*
<details><summary><i>Previously, djoyner (David Joyner) wrote…</i></summary><blockquote>

Since they are separate structures with (apparently) separate lifetimes and identifiers, would it make more sense to surface `/sockets` separately from `/pcbs`? I'd rather see us structure the REST API in a way that more accurately reflects internals than for perceived client convenience. Also keep in mind that a representation of a socket (`/socket/{id}`) could include an optional representation of the associated PCB.
</blockquote></details>

The intent was to provide something which is similar to "netstat -lp".
As implemented I get that behavior with a simple curl call.
If we require 2 separate API calls, then you also need a script to correlate the 2 responses and possibly expose the PCB (pointer/ID) in the socket stats so you can correlate PCB objects and socket objects.
There could also be issues dealing with separate socket/pcb results when doing high rate socket open/close and performance would not be as good when scaling up the number of sockets (because more requests required to get all the info).
So I would prefer to keep the merged socket/pcb REST function.

I was testing a bit more and I think all sockets will have PCB objects associated with them,  so we can always use the PCB address as the unique ID.
All PCBs however, do not have sockets because TCP has a few states to go through after the socket is closed and usually these are the PCBs which are interesting...

The code currently gets TCP/UDP PCB stats only for PCBs registered with the lwip stack because the type of the PCB is known.
If a socket has a PCB but the PCB is not registered with the stack, currently there is no way to get the type of the PCB so TCP/UDP PCB stats are not returned.
Normally this state should only be transient,  but I have seen it a few times while debugging issues.
I guess I can modify the generic_socket class to expose a way to get the type of the PCB so we can return the TCP/UDP PCB stats for those sockets as well.

DerangedMonkeyNinja (approved):
:lgtm:

> Reviewed 10 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/477#-:-MXCPb-q6XzhHzF0OpRI:bnfp4nl)* status: 20 of 29 files reviewed, all discussions resolved (waiting on @DerangedMonkeyNinja)

PR #495, opened 2021-05-24
Add computation Dynamic Results by periods

### LINKED ISSUE #493: OpenPerf calculates threshold results incorrectly
We are trying to configure thresholds (dynamic results) on a memory generator. The threshold is intended to test number of read/write operations per second.
We tried to configure the threshold like this:
```
'config': {
    'buffer_size': 4294967296,
    'pattern': 'random',
    'read_size': 4096,
    'read_threads': 1,
    'reads_per_sec': 100,
    'write_size': 4096,
    'write_threads': 0,
    'writes_per_sec': 0
}
```
Threshold is configured as following:
```
{
    'condition': 'greater',
    'function': 'dxdt',
    'id': 'my0',
    'stat_x': 'read.ops_actual',
    'stat_y': None,
    'value': 0.0
}
```
It is expected that most of the threshold checks will be successful because the generator definitely can perform more than 0 operations per second. But when I look at results I see that about 50% threshold checks failed. It looks like OpenPerf performs threshold checks at every iteration and half of these iterations performed 0 read operations.
```
{
    'id': 'my0',
    'condition_false': 2864,
    'condition_true': 2909,
    ...
}
```

loskutnikov-spirent:
Current implementation of Dynamic Results has some disadvantages:
1. non-constant period of calculation, because of statistics data arrives after each load algorithm spin and Dynamic Results calculates after each spin.
2. zero values in some cases. As consequence of the first peculiarity and concurrent execution some fields of the statistics may be unchanged, thus delta values can be 0.

To resolve this two problems I add computation of Dynamic Results by statically periods. Now, Dynamic Results will be calculated after period of time is elapsed. Default period time is `1s`.
Resolves #493.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/495)

### DIFF
--- a/src/modules/dynamic/spool.hpp
+++ b/src/modules/dynamic/spool.hpp
@@ -234,11 +234,8 @@ template <typename T> double spool<T>::elapsed_periods(const T& stat) const
 {
     auto t = m_extractor(stat, "timestamp").value();
     auto last_t = m_extractor(m_last_stat, "timestamp").value();
-    auto period = std::chrono::duration_cast<std::chrono::nanoseconds>(
-                      m_computation_period)
-                      .count();
 
-    return (t - last_t) / period;
+    return (t - last_t) / m_computation_period.count();
 }
 
 template <typename T>
--- a/src/modules/dynamic/spool.hpp
+++ b/src/modules/dynamic/spool.hpp
@@ -35,12 +35,14 @@ private:
         tdigest tdigest;
     };
 
+    // Constants
+    static constexpr std::chrono::nanoseconds m_computation_period = 1s;
+
     // Attributes
     std::map<std::string, threshold_arg> m_thresholds;
     std::map<std::string, tdigest_arg> m_tdigests;
     extractor m_extractor;
     T m_last_stat;
-    std::chrono::nanoseconds m_computation_period = 1s;
 
 public:
     explicit spool(extractor&& f);
--- a/src/modules/dynamic/spool.hpp
+++ b/src/modules/dynamic/spool.hpp
@@ -12,6 +12,8 @@
 
 namespace openperf::dynamic {
 
+using namespace std::chrono_literals;
+
 template <typename T> class spool
 {
 public:
@@ -38,6 +40,7 @@ private:
     std::map<std::string, tdigest_arg> m_tdigests;
     extractor m_extractor;
     T m_last_stat;
+    std::chrono::nanoseconds m_computation_period = 1s;
 
 public:
     explicit spool(extractor&& f);
@@ -53,6 +56,7 @@ private:
     uint64_t weight(const argument_t&, const T&) const;
     double delta(const argument_t&, const T&) const;
     void argument_check(const argument_t&) const;
+    double elapsed_periods(const T&) const;
 };
 
 //
@@ -158,6 +162,8 @@ template <typename T> void spool<T>::reset()
 
 template <typename T> void spool<T>::add(const T& stat)
 {
+    if (elapsed_periods(stat) < 1) return;
+
     for (auto& th : m_thresholds) {
         auto& data = th.second;
         data.threshold.append(delta(data.argument, stat));
@@ -224,6 +230,17 @@ uint64_t spool<T>::weight(const argument_t& arg, const T& stat) const
     return weight;
 }
 
+template <typename T> double spool<T>::elapsed_periods(const T& stat) const
+{
+    auto t = m_extractor(stat, "timestamp").value();
+    auto last_t = m_extractor(m_last_stat, "timestamp").value();
+    auto period = std::chrono::duration_cast<std::chrono::nanoseconds>(
+                      m_computation_period)
+                      .count();
+
+    return (t - last_t) / period;
+}
+
 template <typename T>
 double spool<T>::delta(const argument_t& arg, const T& stat) const
 {
@@ -238,9 +255,7 @@ double spool<T>::delta(const argument_t& arg, const T& stat) const
         break;
     }
     case argument_t::DXDT: {
-        auto t = m_extractor(stat, "timestamp").value();
-        auto last_t = m_extractor(m_last_stat, "timestamp").value();
-        auto delta_t = t - last_t;
+        auto delta_t = elapsed_periods(stat);
         delta = (delta_t != 0.0) ? delta_x / delta_t : 0.0;
         break;
     }


DerangedMonkeyNinja (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 3 unresolved discussions (waiting on @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 43 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV4d6L6MQCBhyAohNQ:-MaV4d6L6MQCBhyAohNR:bkka3nq) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L43)):*
> ```cpp
>     extractor m_extractor;
>     T m_last_stat;
>     std::chrono::nanoseconds m_computation_period = 1s;
> ```

This should be a `static constexpr` value. If constexpr can't be used because of some weird template reason, then use `static const`.
___
*[src/modules/dynamic/spool.hpp, line 237 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV54nF4UBBr3Y_ak3Q:-MaV54nF4UBBr3Y_ak3R:b66k99c) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L237)):*
> ```cpp
>     auto t = m_extractor(stat, "timestamp").value();
>     auto last_t = m_extractor(m_last_stat, "timestamp").value();
>     auto period = std::chrono::duration_cast<std::chrono::nanoseconds>(
> ```

You don't need to cast this to nanoseconds as it's already a nanoseconds type.  Pretty sure you can just use `m_computation_period.count()` in place of period below.
___
*[src/modules/dynamic/spool.hpp, line 241 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV5aTKC-OgwscQhZDn:-MaV5aTKC-OgwscQhZDo:bsm19eq) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L241)):*
> ```cpp
>                       .count();
> 
>     return (t - last_t) / period;
> ```

Unless t/last_t are floating point numbers, this operation will return an int (period is definitely an int here). Please cast one of these values to double if you want the division to be done as floating point.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 4 unresolved discussions (waiting on @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MaV3wd53MUsny_Ac47u:blemm4e) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*
> ```cpp
>         break;
>     }
>     case argument_t::DXDT: {
> ```

I am confused.  How does this work?
Seems like delta_t would have a different scale factor than before so delta would be different.

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent, @DerangedMonkeyNinja, and @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 43 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV4d6L6MQCBhyAohNQ:-MaW7CIfCHG7R1idAKJO:b-mdngd3) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L43)):*

This should be a `static constexpr` value. If constexpr can't be used because of some weird template reason, then use `static const`.
</blockquote></details>

`m_computation_period` isn't a const and shouldn't be a `const`, because assumed that `period` is a configurable value and can be changed in runtime (I mean runtime of OpenPerf, not generator of course). User might want to use other value for period to get not only smth/sec, but also smth/min or even more exotical smth/100ms or smth/5s.
___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MaWAALcCK4-vjMxuuoq:b-byas4r) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

I am confused.  How does this work?
Seems like delta_t would have a different scale factor than before so delta would be different.
</blockquote></details>

Here OP calculates `delta_t` as number of elapsed periods. One period size is defined by configuration/user. If period is 1s then result will be dx/s (ops/s, for example), if period is 1m result will be dx/min (ops/min). But in general way user can use not only one-unit periods. He can use period of 5s to calculate thresholds every 5s and get dx/5s values, operations per 5s, for example.

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent, @DerangedMonkeyNinja, and @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 241 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV5aTKC-OgwscQhZDn:-MaWJRsh6yRzDcB2YbDZ:bkwknll) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L241)):*

Unless t/last_t are floating point numbers, this operation will return an int (period is definitely an int here). Please cast one of these values to double if you want the division to be done as floating point.
</blockquote></details>

This is not true, result of the expression is double. https://onlinegdb.com/FrxDUM9XO

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/dynamic/spool.hpp, line 237 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV54nF4UBBr3Y_ak3Q:-MaWPg0kBBL_3OXc1QHU:bjcs5dc) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L237)):*

You don't need to cast this to nanoseconds as it's already a nanoseconds type.  Pretty sure you can just use `m_computation_period.count()` in place of period below.
</blockquote></details>

Indeed. Done.

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MaWYCuS5jMNN-vYilSk:b-dday7u) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

So application would need to know configured computation_period to calculate dx/dt?  e.g. if you set period to 5s, then delta is 5x larger?
I guess if it is always 1 then it doesn't need to know...
</blockquote></details>

Application already knows it (`m_computation_period` is an attribute of the class). Suppose the elapsed time is 60s, for 1s period we get 60 periods, for 5s period we get 12 periods. Thus for the same `dx`, for example 240 operations, we get: for the period of 1s is `240 / 60 = 4 op/s` and for the period of 5s is `240 / 12 = 20 op/5s` (20 ops each 5s). Actually this is the same value, because: `20 / 4 == 5 / 1` and `20 op/5s == 4 op/s`. Therefore the answer for your second question is: No, for the larger period delta is smaller, because `delta_t` is not a time in the time units, but rather time expressed in the units of the period.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @DerangedMonkeyNinja and @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MaWkk3q6f70aRllJp8r:b-fiucpo) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

Application already knows it (`m_computation_period` is an attribute of the class). Suppose the elapsed time is 60s, for 1s period we get 60 periods, for 5s period we get 12 periods. Thus for the same `dx`, for example 240 operations, we get: for the period of 1s is `240 / 60 = 4 op/s` and for the period of 5s is `240 / 12 = 20 op/5s` (20 ops each 5s). Actually this is the same value, because: `20 / 4 == 5 / 1` and `20 op/5s == 4 op/s`. Therefore the answer for your second question is: No, for the larger period delta is smaller, because `delta_t` is not a time in the time units, but rather time expressed in the units of the period.
</blockquote></details>

Sorry, I meant application=cloudsure, not application=openperf.
Are you saying that if period = 5s cloudsure would get correct rates without divide by 5?
If cloudsure has to divide by period, then it would be better to do that in openperf.

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MaWq3LhDQc6nTLxlBnS:b-ql48rv) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

Sorry, I meant application=cloudsure, not application=openperf.
Are you saying that if period = 5s cloudsure would get correct rates without divide by 5?
If cloudsure has to divide by period, then it would be better to do that in openperf.
</blockquote></details>

Well it depends on what CS wants to get. If it needs dynamic results with 1s period in that case CS should configure 1s period. If it needs 5s period in that case CS should configure 5s period. I think OP shouldn't convert `dx/dt` values to one-unit period or some fixed period value. One simple thing: user should get what he wants. I think we cannot be sure what user wants exactly: values of 5s periods, or values of 5s periods to get 1s period values later. Of course this value should be configurable through configuration of dynamic results passed on the start of generator. But right now we need working dynamic results with 1s period. I think I can add the configuration capability in another PR.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @DerangedMonkeyNinja and @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MaXEzYk7OyV2LxX5HvF:bqyfcjx) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

Well it depends on what CS wants to get. If it needs dynamic results with 1s period in that case CS should configure 1s period. If it needs 5s period in that case CS should configure 5s period. I think OP shouldn't convert `dx/dt` values to one-unit period or some fixed period value. One simple thing: user should get what he wants. I think we cannot be sure what user wants exactly: values of 5s periods, or values of 5s periods to get 1s period values later. Of course this value should be configurable through configuration of dynamic results passed on the start of generator. But right now we need working dynamic results with 1s period. I think I can add the configuration capability in another PR.
</blockquote></details>

I can't really think of any cases where x/period would make sense (since the value is floating point).
The function is dx/dt which implies (x1-x0)/(t1-t0), so I think it is more consistent to return  x/sec vs x/period.

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MaXMkWx2O9GabLCiHYx:bfr8q8s) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

I can't really think of any cases where x/period would make sense (since the value is floating point).
The function is dx/dt which implies (x1-x0)/(t1-t0), so I think it is more consistent to return  x/sec vs x/period.
</blockquote></details>

The function will imply (x1-x0)/(t1-t0) in both cases. x/sec just a special case of x/period for period=1s. I see no reason to create limitations here.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @cmyers-spirent, @DerangedMonkeyNinja, and @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 43 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV4d6L6MQCBhyAohNQ:-MaYZDLi3u1HJcV3xubS:b-jcsqek) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L43)):*

`m_computation_period` isn't a const and shouldn't be a `const`, because assumed that `period` is a configurable value and can be changed in runtime (I mean runtime of OpenPerf, not generator of course). User might want to use other value for period to get not only smth/sec, but also smth/min or even more exotical smth/100ms or smth/5s.
</blockquote></details>

How does the user change this value?
___
*[src/modules/dynamic/spool.hpp, line 241 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV5aTKC-OgwscQhZDn:-MaYYjFl1Df2r_8n1lyB:b-jvdto) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L241)):*

This is not true, result of the expression is double. https://onlinegdb.com/FrxDUM9XO
</blockquote></details>

Yes, if either `t` or `last_t` are doubles, then the result is a double.  However, I think they are integers. 
If you're using the time sync code to get those timestamps, they should be...

DerangedMonkeyNinja (commented):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 3 unresolved discussions (waiting on @cmyers-spirent and @loskutnikov-spirent)

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 3 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/dynamic/spool.hpp, line 43 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV4d6L6MQCBhyAohNQ:-MaY_Fd_BmV4XHxoIjdm:bocxpif) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L43)):*

How does the user change this value?
</blockquote></details>

I think this value should be configurable through configuration of dynamic results passed on the start of generator. I think I can add the configuration capability in another PR, if it really needed right now.
___
*[src/modules/dynamic/spool.hpp, line 241 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV5aTKC-OgwscQhZDn:-MaY_-fFDwQmH8-FdOX2:b-v2shxq) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L241)):*

Yes, if either `t` or `last_t` are doubles, then the result is a double.  However, I think they are integers. 
If you're using the time sync code to get those timestamps, they should be...
</blockquote></details>

They are `double`, because extractor function has `double` return value...

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 2 unresolved discussions (waiting on @cmyers-spirent and @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 43 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV4d6L6MQCBhyAohNQ:-MaYmjVg4CRd5xA6Rs0w:b-huo2ib) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L43)):*

I think this value should be configurable through configuration of dynamic results passed on the start of generator. I think I can add the configuration capability in another PR, if it really needed right now.
</blockquote></details>

I don't know if it's needed now or not, but until that change is made, please make this static const/constexpr to indicate that it's (currently) unchangeable.
___
*[src/modules/dynamic/spool.hpp, line 241 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV5aTKC-OgwscQhZDn:-MaYmrr_0x_reEK6LroT:bb9yxxv) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L241)):*

They are `double`, because extractor function has `double` return value...
</blockquote></details>

Ahh, ok. I was unaware the extractor was converting the type. Nevermind.

loskutnikov-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/dynamic/spool.hpp, line 43 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV4d6L6MQCBhyAohNQ:-MaYuTYs47aCnOGWi1HT:b-896fix) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L43)):*

I don't know if it's needed now or not, but until that change is made, please make this static const/constexpr to indicate that it's (currently) unchangeable.
</blockquote></details>

Done.

cmyers-spirent (approved):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @DerangedMonkeyNinja and @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-Ma_oW8O5hYrAb1cxX-V:b7rpu3q) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

The function will imply (x1-x0)/(t1-t0) in both cases. x/sec just a special case of x/period for period=1s. I see no reason to create limitations here. 
</blockquote></details>

It makes it harder to use result values.
Since you only support period=1s for now, I guess it is okay for now.

cmyers-spirent (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja)

DerangedMonkeyNinja (approved):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MacWhFK9dJorsnJugH8:b-raba63) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

It makes it harder to use result values.
Since you only support period=1s for now, I guess it is okay for now.
</blockquote></details>

This calculation is just wrong. Just using dimensional analysis, the value returned by `elapsed_periods()` is dimensionless. The `delta_t` value needs to be a unit of seconds, so the correct denominator would just be `last time - current time`.

DerangedMonkeyNinja (changes requested):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 2 unresolved discussions (waiting on @loskutnikov-spirent)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/495#-MacXTAEEQt1yOG4rsQW:-MacXTAEEQt1yOG4rsQX:bepy21d) (no related file):*
I did not mean 

---

DerangedMonkeyNinja (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: all files reviewed, 1 unresolved discussion (waiting on @loskutnikov-spirent)

*[a discussion](https://reviewable.io/reviews/spirent/openperf/495#-MacXTAEEQt1yOG4rsQW:-MacXmb-EXWChmaL46ee:b98geuf) (no related file):*

I did not mean 
</blockquote></details>

Ignore this.

---

DerangedMonkeyNinja (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495#-:-Mad5oZ5BIm-YXYWrvSf:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-Mad5k7k-PbtIuYCJyyt:b-31nt5s) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

This calculation is just wrong. Just using dimensional analysis, the value returned by `elapsed_periods()` is dimensionless. The `delta_t` value needs to be a unit of seconds, so the correct denominator would just be `last time - current time`.
</blockquote></details>

My mistake. Need more coffee.

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/495)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @loskutnikov-spirent)

___
*[src/modules/dynamic/spool.hpp, line 257 at r1](https://reviewable.io/reviews/spirent/openperf/495#-MaV3wd53MUsny_Ac47t:-MaeT-BW-rG5UVdeD2tq:b-lq9eih) ([raw file](https://github.com/spirent/openperf/blob/f8dc28851c1ac2cc1cfb8b18aa26c30bb813ad7e/src/modules/dynamic/spool.hpp#L257)):*

My mistake. Need more coffee.
</blockquote></details>

yeah.  I think dx/dt should always be x/sec regardless of period used.
otherwise, it is confusing to explain/use.

PR #514, opened 2021-06-25
Host sync

DerangedMonkeyNinja:
Add the ability to sync to the host's time-of-day clock.

Just add a system time-source config, e.g.

```
resources:
   /time-sources/system0:
     kind: "system"
```

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/514)


dmorton-spirent (approved):
> Reviewed 15 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/514)* status: 15 of 26 files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent and @DerangedMonkeyNinja)

___
*[src/modules/timesync/source_system.cpp, line 10 at r1](https://reviewable.io/reviews/spirent/openperf/514#-Md3iZiy6Vyfix1N0YWu:-Md3iZiy6Vyfix1N0YWv:b-u65yh8) ([raw file](https://github.com/spirent/openperf/blob/068c769e7abf94cd085e9e3cd7c2c4a61f48555e/src/modules/timesync/source_system.cpp#L10)):*
> ```cpp
> namespace openperf::timesync::source {
> 
> using sys_clock = std::chrono::high_resolution_clock;
> ```

Nit: is it worth adding a warning here that different libc's implement this differently? Thinking that since we now use two different toolchains this could be a source of obscure bugs when comparing ARM and x86 builds.

From cppreference:
`The high_resolution_clock is not implemented consistently across different standard library implementations, and its use should be avoided. It is often just an alias for std::chrono::steady_clock or std::chrono::system_clock, but which one it is depends on the library or configuration. When it is a system_clock, it is not monotonic (e.g., the time can go backwards). For example, for gcc's libstdc++ it is system_clock, for MSVC it is steady_clock, and for clang's libc++ it depends on configuration. 
`

DerangedMonkeyNinja (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/514)* status: 15 of 26 files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent and @dmorton-spirent)

___
*[src/modules/timesync/source_system.cpp, line 10 at r1](https://reviewable.io/reviews/spirent/openperf/514#-Md3iZiy6Vyfix1N0YWu:-Md54Ywx8udUHVykihIQ:b-l3zw06) ([raw file](https://github.com/spirent/openperf/blob/068c769e7abf94cd085e9e3cd7c2c4a61f48555e/src/modules/timesync/source_system.cpp#L10)):*

Nit: is it worth adding a warning here that different libc's implement this differently? Thinking that since we now use two different toolchains this could be a source of obscure bugs when comparing ARM and x86 builds.

From cppreference:
`The high_resolution_clock is not implemented consistently across different standard library implementations, and its use should be avoided. It is often just an alias for std::chrono::steady_clock or std::chrono::system_clock, but which one it is depends on the library or configuration. When it is a system_clock, it is not monotonic (e.g., the time can go backwards). For example, for gcc's libstdc++ it is system_clock, for MSVC it is steady_clock, and for clang's libc++ it depends on configuration. 
`
</blockquote></details>

D'oh! Thank you for pointing that out. I'll switch this to use the `system_clock`. We definitely want to use the clock that is actually getting synced to something!

cmyers-spirent (commented):
> Reviewed 21 of 26 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/514)* status: 25 of 27 files reviewed, all discussions resolved (waiting on @DerangedMonkeyNinja)

dmorton-spirent (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/514#-:-Md5MmQtExvAUUcZb-KI:bg7igkv)* status: 25 of 27 files reviewed, all discussions resolved (waiting on @DerangedMonkeyNinja)

PR #522, opened 2021-08-27
Packetio fixups

DerangedMonkeyNinja:
Various packetio improvements. The highlights are:

* Don't let the worker control DPDK port start/stop, as some port interrogation functions only work when the port is started and don't generate any errors when they return useless info.
* Add driver quirks so that we can filter and ignore known driver problems.
* prefetch header data before parsing packets in all port callbacks.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/522)


DerangedMonkeyNinja (commented):
> Reviewed 1 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/522)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @DerangedMonkeyNinja)

PR #526, opened 2021-09-17
Transmit behavior updates

DerangedMonkeyNinja:
This PR makes two main changes:

1) Use the expected transmit count instead of the actual transmit count to determine when to stop a running packet generator when the user specifies a transmit limit.

2) Add a new packetio command line option, modules.packetio.dpdk.drop-tx-overruns. This option instructs the transmit path to drop packets instead of buffering them when the transmit queue is full.

---
This change is [Reviewable](https://reviewable.io/reviews/spirent/openperf/526)

### DIFF
--- a/api/client/golang/models/packet_generator_flow_counters.go
+++ b/api/client/golang/models/packet_generator_flow_counters.go
@@ -29,6 +29,13 @@ type PacketGeneratorFlowCounters struct {
 	// Minimum: 0
 	OctetsActual *int64 `json:"octets_actual"`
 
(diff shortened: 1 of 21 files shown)


cmyers-spirent (approved):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/526)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @DerangedMonkeyNinja)

PR #528, opened 2021-10-05
modules/timesync: improve name resolution

DerangedMonkeyNinja:
Update the DNS resolution routine in the timesync module to set an upper
bound on DNS lookup requests. Just return the raw socket data if the
request takes too long in order to avoid waiting on the system DNS
timeout, which we have no control over.

### DIFF
--- a/src/modules/timesync/source_ntp.cpp
+++ b/src/modules/timesync/source_ntp.cpp
@@ -1,10 +1,12 @@
 #include <cinttypes>
+#include <future>
 #include <string>
 
 #include <arpa/inet.h>
 
 #include "core/op_log.h"
 #include "core/op_event_loop.hpp"
+#include "core/op_thread.h"
 #include "timesync/api.hpp"
 #include "timesync/clock.hpp"
 #include "timesync/ntp.hpp"
@@ -13,6 +15,9 @@
 namespace openperf::timesync::source {
 
 static constexpr auto poll_period_adjust = 4s;
+
+static constexpr auto dns_query_timeout = 10ms;
+static uint8_t thread_index = 0;
 
 static int handle_ntp_poll(const struct op_event_data* data, void* arg);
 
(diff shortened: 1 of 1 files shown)


dmortondev (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/528)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @DerangedMonkeyNinja and @dmortondev)

___
*[src/modules/timesync/source_ntp.cpp, line 19 at r1](https://reviewable.io/reviews/spirent/openperf/528#-MlHTPts8XYa2V4gtYNA:-MlHTPts8XYa2V4gtYNB:b-dm10t5) ([raw file](https://github.com/spirent/openperf/blob/654cc239438e4b0b8db3ed436f0f96cff40ef186/src/modules/timesync/source_ntp.cpp#L19)):*
> ```cpp
> static constexpr auto poll_period_adjust = 4s;
> 
> static constexpr auto dns_query_timeout = 10ms;
> ```

Nit: Would it be worth trying to surface this to the API? I figure there are use cases where users know there's no DNS and want to bypass the lookup entirely. Conversely, there's probably cases where users know their DNS is slow and would want to wait a bit longer for it to resolve.

DerangedMonkeyNinja (commented):
> *[Reviewable](https://reviewable.io/reviews/spirent/openperf/528)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dmortondev)

___
*[src/modules/timesync/source_ntp.cpp, line 19 at r1](https://reviewable.io/reviews/spirent/openperf/528#-MlHTPts8XYa2V4gtYNA:-MlHVB6a8c64bqGUc8i-:b1280xz) ([raw file](https://github.com/spirent/openperf/blob/654cc239438e4b0b8db3ed436f0f96cff40ef186/src/modules/timesync/source_ntp.cpp#L19)):*
<details><summary><i>Previously, dmortondev (Dan Morton) wrote…</i></summary><blockquote>

Nit: Would it be worth trying to surface this to the API? I figure there are use cases where users know there's no DNS and want to bypass the lookup entirely. Conversely, there's probably cases where users know their DNS is slow and would want to wait a bit longer for it to resolve.
</blockquote></details>

I actually added this async/timeout code to avoid adding an option.  I guess if the value causes problems we can address it.  As is, I doubt most people would even notice.

PR #558, opened 2023-01-18
modules/packetio: DPDK lcore cpuset support

cmyers-spirent:
This change fixes issues when lcores use more than 1 CPU (cpuset).

When an lcore cpuset spans multiple NUMA nodes the lcore is not assigned to a socket ID.  The stack code was using the invalid socket ID from the lcore thread to determine the mempool and was causing a crash.

The stack code was changed to use socket ID 0 when the socket ID is not valid.  Additional code was added to handle cases where the device NUMA affiliation does not match the CPU NUMA node.

The packetio driver init was also made lcore cpuset aware so it removes cpus from the default cpuset based off of the cpuset assigned to the lcore instead of assuming the lcore ID is the CPU ID.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Spirent/openperf/558)

### DIFF
--- a/src/framework/core/op_cpuset_wrapper.cpp
+++ b/src/framework/core/op_cpuset_wrapper.cpp
@@ -184,7 +184,10 @@ int cpuset_set_affinity(const cpuset& set)
 cpuset cpuset_get_affinity()
 {
     auto set = cpuset();
-    op_thread_get_affinity_mask(set.data());
+    if (auto error = op_thread_get_affinity_mask(set.data())) {
+        throw std::runtime_error("Could not retrieve CPU affinity mask: "
+                                 + std::string(strerror(error)));
+    }
     return (set);
 }
 
--- a/src/modules/packet/stack/dpdk/memory.cpp
+++ b/src/modules/packet/stack/dpdk/memory.cpp
@@ -19,8 +19,47 @@ static rte_mempool* memp_mempools[RTE_MAX_NUMA_NODES] = {};
 
 void packet_stack_memp_initialize()
 {
+    struct rte_mempool* default_mempool = nullptr;
+
     for (auto i = 0; i < RTE_MAX_NUMA_NODES; ++i) {
         memp_mempools[i] = openperf::packetio::dpdk::mempool::get_default(i);
+        if (!default_mempool && memp_mempools[i]) {
+            default_mempool = memp_mempools[i];
+        }
+    }
+
+    // Verify default memory pool was found
+    if (!default_mempool) {
+        OP_LOG(OP_LOG_ERROR,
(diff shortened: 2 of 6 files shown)


jmkho-spirent (commented):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Spirent/openperf/558)* status: all files reviewed, 1 unresolved discussion (waiting on @cmyers-spirent)

___
*[`src/modules/packetio/drivers/dpdk/driver.tcc` line 249 at r1](https://reviewable.io/reviews/Spirent/openperf/558#-NMAwdrbCPRQAyMX4hYT:-NMAwdrbCPRQAyMX4hYU:b-1tk9qw) ([raw file](https://github.com/Spirent/openperf/blob/cb51e738e9acef8a6619fda35a7cb67546d64608/src/modules/packetio/drivers/dpdk/driver.tcc#L249)):*
> ```cpp
>                lcore_id,
>                lcore_cpuset.to_string().c_str());
>         if (cpuset_saved == cpuset) {
> ```

For the first iteration of this loop, this condition will always be true. Should it compare against lcore_cpuset instead of cpuset?

cmyers-spirent (commented):
> *[Reviewable](https://reviewable.io/reviews/Spirent/openperf/558)* status: all files reviewed, 1 unresolved discussion (waiting on @jmkho-spirent)

___
*[`src/modules/packetio/drivers/dpdk/driver.tcc` line 249 at r1](https://reviewable.io/reviews/Spirent/openperf/558#-NMAwdrbCPRQAyMX4hYT:-NMBIX_s26DSAoh8Q8k8:b5yaf7a) ([raw file](https://github.com/Spirent/openperf/blob/cb51e738e9acef8a6619fda35a7cb67546d64608/src/modules/packetio/drivers/dpdk/driver.tcc#L249)):*
<details><summary><i>Previously, jmkho-spirent (Jan-Michael Kho) wrote…</i></summary><blockquote>

For the first iteration of this loop, this condition will always be true. Should it compare against lcore_cpuset instead of cpuset?
</blockquote></details>

You are correct.  That was the intent of the code. 
I guess I didn't see the issue because in real usage I don't mix pinned and unpinned lcores.
Thanks!  I wasn't sure if anyone was going to look at openperf PRs.

jmkho-spirent (commented):
:lgtm:

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Spirent/openperf/558#-:-NMC-Tfi0j7UtIy-bx8l:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @cmyers-spirent)

